PR #17, opened 2019-03-10
Consistent order of generated code

rreilink:
In order to facilitate the merging with (regular) Python code generation, I need generated code which is always the same, regardless of Python version (in order to facilitate comparison of the generated code: refactoring of the code generator should not lead to a change in the generated code)

The current gen_mpy.py depends on Python dict iteration order, which is different between Python versions. I modified it to use collections.OrderedDict to get consistent output.

Only the order of methods, functions, attributes etc is changed, so it will not change the functioning of the generated code.

### DIFF
--- a/gen/gen_mpy.py
+++ b/gen/gen_mpy.py
@@ -7,6 +7,7 @@
 # - When converting mp to ptr (and vice versa), verify that types are compatible. Now all pointers are casted to void*.
 
 from __future__ import print_function
+import collections
 import sys
 import struct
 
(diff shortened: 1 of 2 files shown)


embeddedt:
@amirgon Please review this, since it's your code.

amirgon:
@rreilink 
I understand the need for generating the same output from my Micropython binding script and your binding script.  
However, this is only needed for the unification of the Python and Micropython bindings, so we could have a single script to generate them both.   

Before rushing into this unified script, I think we should complete the discussion we started on littlevgl/lvgl#557 and reach agreements and conclusions about how this unified script should look like and how it should work.  
For that purpose I've started a discussion on littlevgl/lvgl#935 ([this post](https://github.com/littlevgl/lvgl/issues/935#issuecomment-472193278)), let's continue there and only after everything is cleared up let's start the implementation.

rreilink:
Thanks @amirgon !

PR #23, opened 2019-04-20
Emscripten patches

embeddedt:
This enables Emscripten to work with the SDL drivers.

### DIFF
--- a/driver/SDL/modSDL.c
+++ b/driver/SDL/modSDL.c
@@ -1,6 +1,9 @@
 #include "../include/common.h"
 #include "SDL_monitor.h"
 #include "SDL_mouse.h"
+#ifdef __EMSCRIPTEN__
+#include <emscripten.h>
+#endif
 
 //////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


driver/SDL/modSDL.c
```diff
@@ -25,11 +29,22 @@ STATIC int tick_thread(void * data)
...
+        mp_sched_schedule((mp_obj_t)&mp_lv_task_handler_obj, mp_const_none);
+        lv_tick_inc(2);
+}
+#endif
 
 STATIC mp_obj_t mp_init_SDL()
 {
     monitor_init();
+#ifdef __EMSCRIPTEN__
+    emscripten_set_main_loop(mp_lv_main_loop, 500, 0);
```
amirgon: 500 FPS, and `lv_tick_inc(2)` must be related right?
What do you think about defining some value and deriving these values from it?

embeddedt: Good idea. I just pushed a commit which defines `LV_TICK_RATE`.

driver/SDL/modSDL.c
```diff
@@ -1,6 +1,14 @@
-#include "../include/common.h"
+[B#include "../include/common.h"
```
amirgon: `[B` typo before `#include`?

embeddedt: Sorry! I didn't compile-test it before pushing.

PR #30, opened 2019-06-26
provides a SDL.run_once() method suitable for use with asyncio

pmp-p:
for keeping track, this is what i used to run on https://pmp-p.github.io/upy_lvgl/index.html port using https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame via emscripten + asyncio loop instead of non deterministic micropython.scheduler function.

The tick value may be wrong on asm.js/wasm platform depending on screen refresh rate because it controls renderings loops intervals (  roughly  1000/ ??? Hz milliseconds ).

### DIFF
--- a/driver/SDL/modSDL.c
+++ b/driver/SDL/modSDL.c
@@ -8,18 +8,37 @@
 /* Defines the LittlevGL tick rate in milliseconds. */
 /* Increasing this value might help with CPU usage at the cost of lower
  * responsiveness. */
-#define LV_TICK_RATE 50
+#define LV_TICK_RATE 16
 
 //////////////////////////////////////////////////////////////////////////////
 
+
+#if MICROPY_ENABLE_SCHEDULER
 STATIC mp_obj_t mp_lv_task_handler(mp_obj_t arg)
-{  
+{
     if (monitor_active()) monitor_sdl_refr_core();
     lv_task_handler();
     return mp_const_none;
 }
-
 STATIC MP_DEFINE_CONST_FUN_OBJ_1(mp_lv_task_handler_obj, mp_lv_task_handler);
+
+#else
+STATIC mp_obj_t
+mp_lv_task_handler(void) {
+
+    if (monitor_active())
+        monitor_sdl_refr_core();
+
+    lv_task_handler();
+    lv_tick_inc(LV_TICK_RATE);
+    return mp_const_none;
+
+}
+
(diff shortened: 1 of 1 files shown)


embeddedt:
> The tick value may be wrong on asm.js/wasm platform

It can't be wrong, otherwise LittlevGL won't behave properly.

Do we really need it to be matching the frame rate? We're not rendering a complex 3D game; we just need a tick roughly every 10 milliseconds.

pmp-p:
as the loop call rate is fixed by the browser maybe add a counter to drop frames ? 1 on 2 would still gives 30 FPS. 
or maybe that should not be a constant and give to run_once() call a parameter to let user decides about refresh  ( like idle / active , iirc the loop go to sleep when changing tabs on recent browsers )
or calculate it for user with the hires timer_ns() eg https://github.com/pmp-p/micropython-ports-wasm/blob/dd7a748d7e3e85a62b0594a678d8a54ad4a41085/modgen/embed.pym#L113

embeddedt:
@pmp-p Can you explain why all of this calculation is necessary? The method I was using (fixed call rate of 10 milliseconds, the *same* way it is done on an embedded system) is already working just fine.

pmp-p:
@embeddedt no i can't, i just hacked my way through modSDL :D i have no idea what ticks do !
i only know that emscripten give  seconds or nanoseconds with hires timer as there was issues at some point calling the performance timer.

embeddedt:
@pmp-p So what is wrong with the method that is already in use?

pmp-p:
well i don't know you said at some point it was 2 or 3 faster and tick value was probably wrong so 16 looked as good as any. but i know that if screen refresh rate changes asyncio loop rates will change too because it not under user control like setTimeout or setInterval

That's up to you to judge if it can be a problem to use a hardcoded value in that particular case.

embeddedt:
@pmp-p I think that a hardcoded value and the use of `setTimeout` is a much smarter way of doing things. It guarantees consistent timing across all browsers.

pmp-p:
i tend to agree with you, as i like simple javascript cause i don't know it.  But emscripten favour https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_set_main_loop and also provides a ready-made emulation of refresh rate with  web workers which seems very attractive.

embeddedt:
> also provides a ready-made emulation of refresh rate with web workers which seems very attractive.

True, but we don't need such advanced functionality for LittlevGL. A consistent tick is good enough.

PR #52, opened 2019-11-23
add tools for png decoding

stepansnigirev:
As discussed in https://github.com/littlevgl/lv_examples/pull/40#discussion_r349786977

I moved array conversion to a viper-decorated function, but not sure if I am doing it right.
Anyways it works on unixport and gets some small speed improvements comparing to non-viper version.


amirgon:
Thanks!

PR #55, opened 2019-11-26
ili9341: add rotation, backlight high/low set, color mode, and invert

endtwist:
I’m working with the [M5Stack](https://m5stack.com) ESP32 device and the ili9341 configuration is a bit different than what was hardcoded into the `ili9341.py` module.

This PR adds a number of options to the initialization of this class that allows configuration of the screen size, rotation mode, backlight (M5Stack requires backlight pin to be pulled high), color mode (RGB vs BGR), and invert mode.

Most of the additions were made based on modifications for the M5Stack in this library: https://github.com/m5stack/M5Stack_TFT_ILI9341/blob/m5stack-default/components/tft/

This code isn’t specific to M5Stack and retains the current default values.

_edit: it looks like PR https://github.com/littlevgl/lv_binding_micropython/pull/54 does something similar, though I tried to implement it in a way that doesn’t specifically pertain to the M5Stack config_

### DIFF
--- a/driver/esp32/ili9341.py
+++ b/driver/esp32/ili9341.py
@@ -23,10 +23,15 @@ import gc
 micropython.alloc_emergency_exception_buf(256)
 # gc.threshold(0x10000) # leave enough room for SPI master TX DMA buffers
 
-class ili9341:
+COLOR_MODE_RGB = 0x00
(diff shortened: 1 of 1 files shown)


amirgon:
Looks good.

PR #57, opened 2019-11-29
fix(ili9341): fix invert kwarg

cfobel:
Prior to this commit, the `ili9341` class `__init__` method would raise
a `KeyError` exception if `invert==True`. The issue was that the
following conditional did not add the required `"data"` key/value pair
to the `init_cmds` list item:

    if invert:
        self.init_cmds.append({'cmd': 0x21})

In this commit, add the required `"data"` key/value pair to the
`init_cmds` list item when `invert==True`.

### DIFF
--- a/driver/esp32/ili9341.py
+++ b/driver/esp32/ili9341.py
@@ -102,7 +102,7 @@ class ili9341:
         ]
 
         if invert:
-            self.init_cmds.append({'cmd': 0x21})
+            self.init_cmds.append({'cmd': 0x21, 'data': bytes([])})
 
         self.init()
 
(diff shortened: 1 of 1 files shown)


amirgon:
@cfobel - instead of adding an empty array, how about testing `data`  key existence, like we do with `delay`, when sending the initialization commands?

https://github.com/littlevgl/lv_binding_micropython/blob/3821abcacd7cd4e9a2586341d194f72be432c3c1/driver/esp32/ili9341.py#L290-L296

cfobel:
@amirgon , thanks for the suggestion. I've added a commit to apply your suggested changes.

amirgon:
Merged. Thanks you!

PR #72, opened 2020-04-01
added drivers for STM32F7DISC board

prolomb:
Hi, 
This is my first PR, so if something is wrong or missing please let me know! :)

Basically this PR adds drivers for STM32F7DISC board.
To integrate the drivers in micropython I did the changes described below. 
I would add these changes in https://github.com/littlevgl/lv_micropython .

> i**n lib/libc/string0.c**
	implemented strncpy for lv_binding_micropython/lvgl/src/lv_objx/lv_chart.c
**in ports/stm32/boards/stm32f7xx_hal_conf_base.h**
	#include "stm32f7xx_hal_ltdc.h"
	#include "stm32f7xx_hal_ltdc_ex.h"
	#include "stm32f7xx_hal_dma2d.h"
	#define HAL_LTDC_MODULE_ENABLED
	#define HAL_DMA2D_MODULE_ENABLED
**in ports/stm32/Makefile**
	ifeq ($(MICROPY_PY_LVGL),1)
	CFLAGS += -DMICROPY_PY_LVGL
	SRC_LIB += $(shell find $(LVGL_BINDING_DIR)/driver/stm32/$(BOARD)/*.c)
	ifeq ($(MCU_SERIES),$(filter $(MCU_SERIES),f7))
	SRC_HAL += $(addprefix $(HAL_DIR)/Src/stm32$(MCU_SERIES)xx_,\
		hal_ltdc.c \
		hal_ltdc_ex.c \
		hal_dma2d.c \
		)
	endif
	endif
**in ports/stm32/boards/STM32F7DISC/mpconfigboard.mk**
	MICROPY_PY_LVGL = 1
**in py/py.mk**
	#LittlevGL
	ifeq ($(MICROPY_PY_LVGL),1)
	LVGL_BINDING_DIR = $(TOP)/lib/lv_binding_micropython
	LVGL_DIR = $(LVGL_BINDING_DIR)/lvgl
	LVGL_GENERIC_DRV_DIR = $(LVGL_BINDING_DIR)/driver/generic
	INC += -I$(LVGL_BINDING_DIR)
	ALL_LVGL_SRC = $(shell find $(LVGL_DIR) -type f -name '*.h') $(LVGL_BINDING_DIR)/lv_conf.h
	LVGL_PP = $(BUILD)/lvgl/lvgl.pp.c
	LVGL_MPY = $(BUILD)/lvgl/lv_mpy.c
	LVGL_MPY_METADATA = $(BUILD)/lvgl/lv_mpy.json
	QSTR_GLOBAL_DEPENDENCIES += $(LVGL_MPY)
	CFLAGS_MOD += $(LV_CFLAGS) 
	$(LVGL_MPY): $(ALL_LVGL_SRC) $(LVGL_BINDING_DIR)/gen/gen_mpy.py 
		$(ECHO) "LVGL-GEN $@"
		$(Q)mkdir -p $(dir $@)
		$(Q)$(CPP) $(LV_CFLAGS) -I $(LVGL_BINDING_DIR)/pycparser/utils/fake_libc_include $(INC) $(LVGL_DIR)/lvgl.h > $(LVGL_PP)
		$(Q)$(PYTHON) $(LVGL_BINDING_DIR)/gen/gen_mpy.py -M lvgl -MP lv -MD $(LVGL_MPY_METADATA) -E $(LVGL_PP) $(LVGL_DIR)/lvgl.h > $@
	CFLAGS_MOD += -Wno-unused-function
	SRC_MOD += $(subst $(TOP)/,,$(shell find $(LVGL_DIR)/src $(LVGL_GENERIC_DRV_DIR) -type f -name "*.c") $(LVGL_MPY))
	endif
**in ports/stm32/mpconfigport.h**
	#if MICROPY_PY_LVGL
	extern const struct _mp_obj_module_t mp_module_lvgl;
	#include "lib/lv_binding_micropython/lvgl/src/lv_misc/lv_gc.h"
	#define MICROPY_PORT_LVGL_DEF \
	    { MP_OBJ_NEW_QSTR(MP_QSTR_lvgl), (mp_obj_t)&mp_module_lvgl },
	#else
	#define MICROPY_PORT_LVGL_DEF
	#define LV_ROOTS
	#endif
```

### CODE REVIEW

**Comment by amirgon** [comment]
Thank you for this PR @prolomb !  
I think it could be a valuable contribution to LittlevGL. Several people have already asked for STM32 Micropython drivers in the past. 

I have no experience with STM32 (I'm using mainly ESP32), so perhaps @embeddedt or @kisvegabor who have more experience with STM32 could also review your code.

Here are a few comments/questions about this PR:

- I see you added many files under `STM32F7DISC` directory. Are all these files specific to this version of STM32?  
If some of these files could be used by other (non F7DISC) variations of STM32, it might be better to arrange these files differently.
- Are the screen/touch connections of  `STM32F7DISC` hard wired or configurable?  
In case of ESP32 driver, for example, all pins (miso, mosi, clk, cs, dc, rst, power, backlight) [are configurable on runtime](https://github.com/littlevgl/lv_binding_micropython/blob/49e5b6562f566995030d241655be2c02f4154a54/driver/esp32/ili9341.py#L48-L52), as well as clock speed, rotation, color mode etc. This means that the user can set them in his Python script instead of rebuilding the firmware.  
On your case, very few parameters (only `w` and `h` as far as I can see) seem to be configurable without firmware rebuild.
- On `modlcd.c` you create a timer and call both `lv_tick_inc` and `lv_task_handler` on every tick one after the other.  I don't think it should work that way, these are two separate functions for a reason.  
`lv_tick_inc` is very short and can be called from interrupt context. It should be called at precisely the right time, without missing ticks, to maintain correct timing in lvgl.  
`lv_task_handler` on the other hand can take longer and cannot be called from interrupt context. But it doesn't have to be called in an "accurate" manner.  
When you call them together, if `lv_task_handler` takes too long (and overflows to the next time slot), it would cause `lv_tick_inc` not to be called at the correct timing.

**Comment by embeddedt** [comment]
> Are the screen/touch connections of `STM32F7DISC` hard wired or configurable?

The screen is part of the development board and is not removable, so I would say that it's hard-wired.

I haven't looked at the code in detail or tested it yet.

**Comment by prolomb** [comment]
Hi @amirgon 
Regarding your comments:
- Most of the files are taken from STM32Cube_FW_F7_V1.16.0 package provided by STmicroelectronics. We may eventually use a similar arrangement of the file tree like this:

> stm32/Components/Common/ts.h
stm32/Components/ft5336/ft5336.h
stm32/Components/ft5336/ft5336.c
stm32/STM32F7DISC/stm32746g_discovery.h
stm32/STM32F7DISC/stm32746g_discovery_ts.c
stm32/STM32F7DISC/stm32746g_discovery_ts.h
stm32/STM32F7DISC/modlcd.c
stm32/STM32F7DISC/modts.c
- I aggree with @embeddedt. 
- Understood! Do you suggest to use mp_sched_schedule for lv_task_handler?

I forgot to say that the driver supports 16bit and 32bit color depth

**Comment by prolomb** [comment]
@kisvegabor 
I was not aware of that port, thank you! 
btw a couple of points sounds strange to me in [tft.c](https://github.com/littlevgl/lv_port_stm32f746_disco_sw4stm32/blob/master/hal_stm_lvgl/tft/tft.c)

1. I see that DMA was used in `ex_disp_flush` instead of DMA2D. What is the reason behind this? I think the right hardware for the job is DMA2D not DMA which would be better to keep free for other stuff. 
2. Data cache is never flushed (Probably not used?)
3. Why in `gpu_mem_fill` the `HAL_DMA2D_BlendingStart` is used instead of `HAL_DMA2D_Start`?

**Comment by amirgon** [comment]
> * Understood! Do you suggest to use:
> 
> ```
> STATIC mp_obj_t mp_lv_task_handler(mp_obj_t arg)
> {
>     lv_task_handler();
>     return mp_const_none;
> }
> 
> STATIC MP_DEFINE_CONST_FUN_OBJ_1(mp_lv_task_handler_obj, mp_lv_task_handler);
> 
> STATIC mp_obj_t mp_lv_task_handler(mp_obj_t arg)
> {
>     lv_tick_inc(LV_TICK_RATE);
>     mp_sched_schedule((mp_obj_t)&mp_lv_task_handler_obj, mp_const_none);
> }
> ```
> 
> or maybe it could be better to rely on [lvstm32.py](https://github.com/littlevgl/lv_micropython/blob/master/ports/stm32/modules/lvstm32.py)?

@prolomb These would actually do the same thing!
You can create timers and schedule tasks on both C and Micropython.
Personally, I usually prefer doing everything in Python except from cases where performance is important.  
For example, ili9341 display driver for ESP32 is mostly written in python, while the performance critical parts are implemented in C.  

Regardless of whether in Python or C, I suggest keeping the module that calls `lv_tick_inc` and `lv_task_handler` - **generic for STM32** and not specific to a deveopment board model.

**Comment by kisvegabor** [comment]
> I see that DMA was used in ex_disp_flush instead of DMA2D. What is the reason behind this? I think the right hardware for the job is DMA2D not DMA which would be better to keep free for other stuff.

Let's say you use 2 display buffers in lvgl and  `flush_cb` uses DMA2D and you use GPU too. In this case while the first buffer isbeing flushed by DMA2D, lvgl can't use GPU while drawing into the second buffer because DMA2D is still busy with flushing. It can be solved by using DMA in `flush_cb`. It's a little bit hacky solution so if you a better idea I'd be happy to hear it.

> Data cache is never flushed (Probably not used?)

The cache is not enabled in the `master` branch. We played with this project while developing v7 and updated the drivers too. See the [dev-7.0](https://github.com/littlevgl/lv_port_stm32f746_disco_sw4stm32/blob/dev-7.0/hal_stm_lvgl/tft/tft.c) branch.

> Why in gpu_mem_fill the HAL_DMA2D_BlendingStart is used instead of HAL_DMA2D_Start?

True. It still needs to be fixed in `dev-7.0` too.

**Comment by prolomb** [comment]
@kisvegabor 
If the point here is to optimize for speed i would prefer to use the 2 screen-sized buffers approach rather than using this kind of trick to optimize the 2 non-screen-sized buffers approach.
On the other hand, RAM should not be an issue on this board (~8MB) while I'm not convinced that you actually optimize for speed using DMA because you have concurrent access on the bus from DMA and DMA2D.
Thus, my idea is to reconfigure LTDC within flush function when a full screen buffer is provided and use DMA2D for copying in the other cases.

@amirgon 
I aggree with you. Removing the timer part in `modlcd.c` and using `lvstm32.py` would probably be cleaner. Maybe `lvstm32.py` should be moved from `ports/stm32/modules ` to `lib/lv_bindings_micropython/driver/stm32`

**Comment by amirgon** [comment]
> Maybe `lvstm32.py` should be moved from `ports/stm32/modules ` to `lib/lv_bindings_micropython/driver/stm32`

@prolomb - Good idea.

**Comment by kisvegabor** [comment]
@prolomb 
> If the point here is to optimize for speed i would prefer to use the 2 screen-sized buffers approach rather than using this kind of trick to optimize the 2 non-screen-sized buffers approach.

I've found that using 2 screen-sized buffers in external RAM is slower. I suppose it happens because during rendering every pixel is (over)written and read multiple times (e.g. background, button, text). In the end, accessing the slower RAM multiple times for every pixel took more time than coping a buffer to it (in this case 1 pixel is written only once).

BTW, it seems you know the STM hardware very well. On STM32F769 Discovery I've tried to get an interrupt on VSYNC. I've used `HAL_LTDC_ProgramLineEvent(hltdc, 0);` and called it again in the interrupt callback but didn't get interrupt. Do you have experience with it? If so can you open an issue for it in the [stm32f769](https://github.com/littlevgl/lv_port_stm32f769_disco_sw4stm32/) repo?

**Comment by prolomb** [comment]
@kisvegabor 

> 've found that using 2 screen-sized buffers in external RAM is slower. I suppose it happens because during rendering every pixel is (over)written and read multiple times (e.g. background, button, text). In the end, accessing the slower RAM multiple times for every pixel took more time than coping a buffer to it (in this case 1 pixel is written only once).

I think that in micropython when we dynamically allocate buffers it is not possible to chose the RAM region thus unfortunately this reduction in performance is unavoidable no matter the selected buffer strategy!

> Do you have experience with it?

No. However if the callback is never called I can suggest you to check NVIC in your LTDC init (probably already done)
```
HAL_NVIC_SetPriority(LTDC_IRQn, 9, 3);
HAL_NVIC_EnableIRQ(LTDC_IRQn);
```

**Comment by amirgon** [comment]
> I think that in micropython when we dynamically allocate buffers it is not possible to chose the RAM region thus unfortunately this reduction in performance is unavoidable no matter the selected buffer strategy!

@prolomb - Could you elaborate on that? You can still allocate the RAM in C code, using ST libraries. You don't have to allocate it in Micropython.  
In such case, would you still have trouble choosing the RAM region?

**Comment by prolomb** [comment]
Hi @amirgon 

> Could you elaborate on that? You can still allocate the RAM in C code, using ST libraries. You don't have to allocate it in Micropython.
In such case, would you still have trouble choosing the RAM region?

The only way I know to put the buffer on MCU RAM is to declare as global variable. I think this is dirty solution since you would waste that RAM in mp build when you don't use lvgl.
BTW, MCU has 340KB of RAM so even without mp environment it would be impossible to allocate one frame buffer in 32bit color depth (480×272×4 = 522K) or two buffers in 16bit color depth.

**Comment by kisvegabor** [comment]
Wouldn't it be possible to simply pass the frame buffers address (in external RAM) to display driver from mp and simply allocate smaller display buffer for lvgl?

**Comment by prolomb** [comment]
@kisvegabor 

> and allocate smaller display buffer for lvgl?

Do you mean to allocate within the driver (e.g. modlcd.c)? This would persist ... although with less impact:

> I think this is dirty solution since you would waste that RAM in mp build when you don't use lvgl.

**Comment by kisvegabor** [comment]
I mean something like [here](https://github.com/littlevgl/lv_binding_micropython/blob/master/examples/example1.py). So allocate lvgl buffers in mp. 
(I'm not a micropython expert, so let me know if I'm saying nonsense things.)

**Comment by prolomb** [comment]
> I mean something like here. So allocate lvgl buffers in mp.
This is what I've done actually! :)
The point is that probably `bytearray(480*10)` uses external RAM on STM32F7DISC

**Comment by kisvegabor** [comment]
Ah, I see. Another option would be to use `lv_mem_alloc` but lvgl uses mp's alloc in the binding so you might end up allocation in ext RAM.

What about using normal C `malloc` in the driver?

**Comment by prolomb** [comment]
@embeddedt 
[Here](https://github.com/prolomb/micropython) you can find the mp build I'm using to test STM32F7DISC. It includes the forked [lv_bindings](https://github.com/prolomb/lv_binding_micropython) of this PR.
The mp branch is lvgl_driver_STM32F7DISC
An example here [lvgl_stm32.py ](https://github.com/prolomb/micropython/blob/lvgl_driver_STM32F7DISC/examples/lvgl_stm32.py)

File: driver/stm32/lvstm32.py [hunk 1]
```diff
@@ -19,7 +19,11 @@ def task_handler(self, _):
         lv.task_handler()
 
     def timer_cb(self, t):
-        lv.tick_inc(self.delay)
-        # Passing self.task_handler would cause allocation.
-        micropython.schedule(self.task_handler_ref, 0)
+	try:
```
amirgon, line 22: What exceptions are you catching here?  
I don't think neither `lv.tick_inc` nor `micropython.schedule` are expected to throw exceptions.  

btw there are also some problems with your indentation.

prolomb: Hi @amirgon 
I see this
 `>>> import lvgl_stm32
uncaught exception in Timer(4) interrupt handler
RuntimeError: `

sorry for indentation. I'm using gedit or CubeMXIDE that use tab instad of spaces... How can I easily fix this?

amirgon: I'm not sure about the editors, I'm using Vim. Just copy the spaces from other lines.  

About the exception - usually it comes with a traceback with more information.  
If you don't see any traceback, please try adding the line  
```python
micropython.alloc_emergency_exception_buf(100)
```
right after the imports, and see if you get the traceback.

prolomb: When calling `micropython.schedule(self.task_handler_ref, 0)` the exception was `schedule queue full`. I found the issue was due to long processing in modlcd now fixed.
I restored slvstm32.py

prolomb:
@kisvegabor @amirgon 

According to received feedback I updated the drivers mainly adding support to 2 screen-sized buffers and Improving DMA2D handling

> What about using normal C malloc in the driver?

In mp we have to use m_malloc & co as stated [here](https://github.com/micropython/micropython/blob/master/CODECONVENTIONS.md) 
However i found that DTCM RAM (64kB) is not used at all in micropython. Thus I implemented **bytarray** in  [modrk043fn48h](https://github.com/littlevgl/lv_binding_micropython/pull/72/commits/4a83f4fa7417e616f90d4394e324ebef78b71c89#diff-3f5e8495bcbc34b71e90d9a44d33f1e5) which allocates small buffers on DTCM improving speed. 
**bytearray** Allocate screen-sized framebuffers on extRAM with 1KB alignment to improve LTDC [performance](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&ved=2ahUKEwjyq4iN-OzoAhXM_KQKHaptBgkQFjABegQIBRAB&url=https%3A%2F%2Fwww.st.com%2Fresource%2Fen%2Fapplication_note%2Fdm00287603-lcdtft-display-controller-ltdc-on-stm32-mcus-stmicroelectronics.pdf&usg=AOvVaw36OQ5QvrgI_385dK5QrKza)

amirgon:
@prolomb I noticed you changed the examples to support stm32 and added a new example.
That's a good thing! but I have a few comments.

Your new example seems stm32 specific, I think that examples should be general enough to run on all platforms.

A way to achieve this is to import [display_driver.py](https://github.com/littlevgl/lv_binding_micropython/blob/master/lib/display_driver.py) on each example, instead of initializing the drivers on each example.  
Today the examples contain the driver initialization in them duplicated across examples. A better approach would be to write it once (on display_driver.py) and import it on the examples code.
On `lv_micropython` it would be a good idea to freeze display_driver.py, such that it is always available.  

An open question is how to select the display driver when more than one option is available on the same platform. While we know what platform the user is using, we don't yet have an automatic way to detect which display type the user is using.

amirgon:
> Thus I implemented **bytarray** in [modrk043fn48h](https://github.com/littlevgl/lv_binding_micropython/pull/72/commits/4a83f4fa7417e616f90d4394e324ebef78b71c89#diff-3f5e8495bcbc34b71e90d9a44d33f1e5) which allocates small buffers on DTCM improving speed.

@prolomb What if some other driver would want to use DTCM?   
Is it possible to allocate it dynamically, only when `mp_rk043fn48h_bytearray` is actually called? 
Currently it's allocated statically, so it would be wasted when the driver is not used.

prolomb:
@amirgon 

> Your new example seems stm32 specific, I think that examples should be general enough to run on all platforms.

I Aggree with you but not scared about that since example2 works only with SDL!

> Today the examples contain the driver initialization in them duplicated across examples.

This is a more general discussion. I think this PR is not right place to discuss about that!
I adhere with current convention in updating the examples.

> Is it possible to allocate it dynamically, only when mp_rk043fn48h_bytearray is actually called?
Currently it's allocated statically, so it would be wasted when the driver is not used.

I know that as written in a previous post. However that was the only way I know at the moment!

lv_conf.h
```diff
@@ -9,14 +9,15 @@
 /* clang-format off */
 
 #include <stdint.h>
+#define mp_instance_cast_to_native_base mp_obj_cast_to_native_base
```
embeddedt, line 12: I don't see this being used anywhere else in the diff; is there a reason why it's here?

lv_conf.h
```diff
@@ -55,7 +56,7 @@
 /* Dot Per Inch: used to initialize default sizes.
  * E.g. a button with width = LV_DPI / 2 -> half inch wide
  * (Not so important, you can adjust it to modify default sizes and spaces)*/
-#define LV_DPI              100     /*[px]*/
+#define LV_DPI              60     /*[px]*/
```
embeddedt, line 59: This affects all platforms, as far as I know. Can we leave it at 100, or at least condition the change for STM32?

lv_conf.h
prolomb: Yes. It is needed to compile with latest micropython where *instance* was replaced with *obj* 
To avoid cross editors issues I formatted all examples with tools/codeformat.py

amirgon:
> > Is it possible to allocate it dynamically, only when mp_rk043fn48h_bytearray is actually called?
> > Currently it's allocated statically, so it would be wasted when the driver is not used.
> 
> I know that as written in a previous post. However that was the only way I know at the moment!

@prolomb I think [flashbdev](https://github.com/micropython/micropython/blob/f6375ac3ebac28656a0a757952d32c265b1ba7aa/ports/stm32/flashbdev.c#L91) uses dtcm.
Wouldn't your driver and flashbdev corrupt each other memory?

lv_conf.h
prolomb: Yes we can leave 100.  On my screen the style was better with 60

lv_conf.h
```diff
@@ -9,14 +9,15 @@
...
+#define mp_instance_cast_to_native_base mp_obj_cast_to_native_base
 
 /*====================
    Graphical settings
  *====================*/
 
 /* Maximal horizontal and vertical resolution to support by the library.*/
 #define LV_HOR_RES_MAX          (480)
-#define LV_VER_RES_MAX          (320)
+#define LV_VER_RES_MAX          (272)	//(320)
```
amirgon, line 20: This change is specific to your driver

prolomb: Correct! i missed it.

lv_conf.h
embeddedt: @amirgon Should `gen_mpy.py` be updated?

prolomb:
> I think flashbdev uses dtcm.
Wouldn't your driver and flashbdev corrupt each other memory?

You are right! The issue comes from the fact that flashbdev.c uses the row address so no compile error was generated and I experienced no issue because I'm using SD card.
So definitely the buffer should be statically allocated on SRAM. Any other idea?

lv_conf.h
prolomb: Honestly I don't know how! May you help me?

lv_conf.h
prolomb: I just checked the examples with 100 and are quite terrible :)

lv_conf.h
amirgon: @embeddedt `gen_mpy.py` is aligned to last official Micropython version.
I think @prolomb wants the code to be compatible with a more recent (unofficial) version, so this define is a workaround.  
After a new Micropython version is out, I'll be able to fix `gen_mpy.py` (and some other drivers)  and remove this define.
For now, I don't think any change is needed on `gen_mpy.py`.

embeddedt: Okay; that's fine.

prolomb:
@amirgon and @kisvegabor 
On STM32F7DISC seems that mp heap is on extRAM.
Do you see any issue using newlibc malloc?

prolomb:
@amirgon 
Is it possible to change default DPI at runtime?
I see that objects are really too large on STM32F7DISC with LV_DPI 100

kisvegabor:
> Is it possible to change default DPI at runtime?

I will be possible in v7.

amirgon:
> Do you see any issue using newlibc malloc?

@prolomb Would your code and Micropython core use the same malloc functions? Or do you mean they will use different heaps?

prolomb:
@amirgon 
They use different heap.
The mp heap is on extRAM while heap for malloc and co. is on SRAM

amirgon:
> They use different heap.
> The mp heap is on extRAM while heap for malloc and co. is on SRAM

So the question is - Is it ok to use SRAM with Micropython?
That probably depends not only on Micropython core, but also on drivers and external libraries.
What if some driver/library use SRAM at the same time? If they both use the same  malloc function that would probably be ok, but if they are independent that could cause trouble (like with DTCM).

prolomb:
@amirgon 
We should come to a conclusion on this point :)

Given that:
1. DTCM can't be used (since cache in case of  files writing on flash).
2. m_malloc allocates on extRAM and slows down the refresh (as stated by @kisvegabor)

The only remaining place to put video buffers is SRAM.
There are two options now:
- Allocate buffers statically (this would waste memory in case not using the driver)
- Allocate buffers dynamically

You wrote
> Is it possible to allocate it dynamically, only when mp_rk043fn48h_bytearray is actually called?
Currently it's allocated statically, so it would be wasted when the driver is not used.

Consequently, I added newlibc since the implementation of malloc is not natively provided in mp (or at least I was not able to find it).

How do you suggest to go further?

1. Accept performance reduction and use extRAM with m_malloc
2. Leave newlibc malloc
3. Allocate statically
4. ...

embeddedt:
I think the only piece of code that should deal with SRAM directly (i.e. without an existing pointer) is the heap implementation. Everything else should either declare a variable and use a linker section, or make use of the heap. If that condition isn't met, it becomes impossible to trace what code is using what memory (and also leads to bugs when two things share the same memory).

Personally, it seems strange to me to add a separate SRAM heap purely for this driver. I would rather just allocate the buffer statically and provide a way for the user to disable the driver if they don't want that much SRAM being used. But that's just my opinion. :slightly_smiling_face:

amirgon:
> How do you suggest to go further?
> 
> 1. Accept performance reduction and use extRAM with m_malloc
> 2. Leave newlibc malloc
> 3. Allocate statically
> 4. ...

@prolomb - I agree with @embeddedt that it makes no sense to add a heap manager for allocating RAM just for your driver.  
The whole idea of a heap manager is that all components are aware of it and using it.  
Multiple heap managers that use the same hardware resource and not aware of each other would just corrupt each other memory.

So my opinion is:
- If there is a way to use a heap manager which is **shared** with other Micropython components (drivers, libraries) - that's the best way.  
I don't know the answer to this, and the right place to ask is [Micropython forums](https://forum.micropython.org/).  
It's a general question, not specific to lvgl - **how to use stm32 SRAM with Micropython** when SRAM could be shared between multiple libraries and drivers.
- If there is no "standard" way to allocate SRAM dynamically, then allocate it statically.  
I suggest you make the size of the allocated SRAM configurable though a macro in `lv_conf.h`, and give it a reasonable default.

embeddedt:
@prolomb It would be great if we could get this merged before the 7.0 release, as it would be the first STM32 board to be supported out of the box with LittlevGL+MicroPython. Have you had a chance to look at @amirgon's suggestions? Don't worry if you haven't; not trying to put pressure on you. :slightly_smiling_face:

prolomb:
Hi @embeddedt 
Unfortunately had no time to work on that neither ask mp forum. I'm quite busy at the moment!
If you agree I could make a clean up commit simply removing the malloc part.
In future revisions we could work on optimizations

amirgon:
> If you agree I could make a clean up commit simply removing the malloc part.

That's fine with me (temporary static allocation).  
I would very much like to see this PR merged.

@embeddedt, @kisvegabor - did you complete the review of this PR from your side?

prolomb:
@amirgon 

> Could you make the static buffer size configurable through a macro in lv_conf.h?

I could but I would prefer the other way for cleanliness: let user allocate buffers via `bytearray` in python code. Is this acceptable for you? (Please note that all graphic objects would remain allocated via m_malloc() as defined in lv_conf.h)

> Would you have time to submit a PR to lv_micropython with the rest of the changes?

Should not be time consuming but cannot work on that until end of next week. In the meantime there is https://github.com/prolomb/micropython with all changes

examples/advanced_demo.py
```diff
@@ -176,29 +193,30 @@ def on_slider_changed(obj=None, event=-1):
...
+
         SDL.init()
 
         # Register SDL display driver.
 
         disp_buf1 = lv.disp_buf_t()
-        buf1_1 = bytes(480*10)
-        lv.disp_buf_init(disp_buf1,buf1_1, None, len(buf1_1)//4)
+        buf1_1 = bytes(480 * 10)
+        lv.disp_buf_init(disp_buf1, buf1_1, None, len(buf1_1) // 4)
```
embeddedt, line 219: This should also be changed to `lv.color_t.SIZE`, right?

embeddedt:
In general, it looks good. I'll test it on my F746 as soon as I can.

examples/advanced_demo.py
prolomb: Correct.

embeddedt:
It runs on my F746, but there is a problem. After some time of running (I haven't been able to measure yet), the screen stops refreshing. If I call `lv.refr_now(None)` from the REPL, it does refresh once, so MicroPython is still working. This might be an issue with the `lvstm32` module.

amirgon:
> I could but I would prefer the other way for cleanliness: let user allocate buffers via `bytearray` in python code. Is this acceptable for you? (Please note that all graphic objects would remain allocated via m_malloc() as defined in lv_conf.h)

@prolomb That's fine with me. You said that performance is not acceptable with external RAM, so I thought you would prefer going with static SRAM allocation.

> Should not be time consuming but cannot work on that until end of next week. In the meantime there is https://github.com/prolomb/micropython with all changes

Whenever you find time for this it'd be great. Thank you.

> This might be an issue with the `lvstm32` module.

@embeddedt [lvstm32 is very simple](https://github.com/littlevgl/lv_micropython/blob/master/ports/stm32/modules/lvstm32.py), I'm not sure what can go wrong there. It's a timer that schedules Micropython task. It's based on example code [from Micropython docs](https://docs.micropython.org/en/v1.12/reference/isr_rules.html). It does involve interrupt handling so maybe something went wrong there.

If the problem is really in lvstm32, it could be very easy to implement it in C. [On ESP32 I implemented it on C](https://github.com/littlevgl/lv_binding_micropython/blob/master/driver/esp32/modlvesp32.c) and it does exactly the same thing.

embeddedt:
I tried it again. This time interaction with objects, etc. continues to work as expected, but the preloader doesn't spin. My guess is that the animation isn't running for some reason, but then how is the screen still refreshing? :confused:

amirgon:
> My guess is that the animation isn't running for some reason, but then how is the screen still refreshing? 😕

Maybe `lv_task_handler` gets called but `lv_tick_inc` doesn't for some reason? Is there a way to make sure that `sys_time` is advancing?

embeddedt:
> Is there a way to make sure that `sys_time` is advancing?

Yes, by using `lv.tick_get()`.

I tried that and found that it is advancing.

prolomb:
@embeddedt 
On my board it run for hours with no issue. Some question:
Are you using https://github.com/prolomb/micropython?
Are you debugging example3?
Did you remove the malloc? 
Does this issue appear all the time you run the example?
Have you tried other examples?
May you eventually provide the binary so that I can try it on my board?

embeddedt:
> Are you using https://github.com/prolomb/micropython?

Yes. I made two changes. One was to checkout `driver_STM32F7DISC` on `lv_binding_micropython`; the other was removing `malloc`.

> Are you debugging example3?

Yes.

> Did you remove the malloc?

Yes.

```diff
diff --git a/driver/stm32/STM32F7DISC/modrk043fn48h.c b/driver/stm32/STM32F7DISC/modrk043fn48h.c
index 1c83a70..4348fb5 100644
--- a/driver/stm32/STM32F7DISC/modrk043fn48h.c
+++ b/driver/stm32/STM32F7DISC/modrk043fn48h.c
@@ -34,9 +34,10 @@ volatile bool dma2d_pend = false;           // flag of DMA2D pending operation
 static bool config_ltdc(void);
 static bool config_dma2d(void);
 
+
 STATIC mp_obj_t mp_rk043fn48h_bytearray(mp_obj_t n_obj) {
     size_t n = (size_t)mp_obj_get_int(n_obj);
-    void *p = malloc(n);         // try allocation on SRAM
+    void *p = NULL; //malloc(n);         // try allocation on SRAM
     if (p == NULL) {
         // allocation on extRAM with 1KB alignment
         p = m_malloc(n + 1024);
```
> Does this issue appear all the time you run the example?

Yes.

> Have you tried other examples?

Not yet.

> May you eventually provide the binary so that I can try it on my board?

[firmware.hex.zip](https://github.com/littlevgl/lv_binding_micropython/files/4580578/firmware.hex.zip)

prolomb:
Hi @embeddedt 
I just tried the firmware but was not not able to reproduce the issue!
I use SD card and run import example3 
I can write via keyboard and go right/left all times i want and the preloader keep spinning
Does the issue appear after seconds from startup?

embeddedt:
That's weird! No; it seems to appear immediately for me. I didn't run the example from the SD card; I plugged a USB cable into the full-speed USB device port (not the ST-link one) and copied the script to the board that way.

Let me try it again and see if I can still make it happen.

embeddedt:
These are the commands I run:

```sh
st-flash --format ihex write firmware.hex
# click PYBFLASH in Files so it mounts
cp ../../lib/lv_binding_micropython/examples/example3.py /media/<user>/PYBFLASH/main.py
sudo umount /media/<user>/PYBFLASH
# reset the board
```

Once LittlevGL appears on the screen, I press the right arrow button and the preloader appears without spinning.

prolomb:
Yes, now I was able to reproduce the issue!
It appears when the example run as main.py
May you try this for xcheck?
Upload the example3 as is on the board (remove main.py) and from REPL type

>  import example3

embeddedt:
It works with your method. I wonder what's special about `main.py`.

prolomb:
may we exclude this is an issue of the driver itself?

embeddedt:
I suppose, given that in both of the two cases I've seen the screen driver continues functioning normally. The issue seems to be related to task scheduling (either at the LittlevGL or MicroPython level).

kisvegabor:
Sorry, for the dumb question but where is the `example3`?

embeddedt:
@kisvegabor It's in `lv_binding_micropython/examples`.

kisvegabor:
@embeddedt 
I've checked [there](https://github.com/littlevgl/lv_binding_micropython/tree/master/examples) but there are only example1 and 2.

prolomb:
Hi @kisvegabor
Here is the example3 https://github.com/prolomb/lv_binding_micropython/tree/driver_STM32F7DISC/examples

kisvegabor:
Ah, thanks.
I don't see anything suspicious. However, I've already faced something similar. If the drawing takes a lot of time it doesn't let the animations running. Might it happen in your case? 
I've already fixed it in v7 by increasing the priority of the animation task.

prolomb:
But the drawing time within the driver should be the same either we run as main.py or directly example3.py
I was wandering if during poweron when we run main.py something else within mp environment is slowing down execution and this is seen by lvgl as slow drawing

embeddedt:
It makes sense to some extent. `main.py` is run immediately, whereas `example3.py` has to be launched by typing `import example3` on the REPL. If startup routines get run alongside `main.py`, it might affect the drawing time. With `example3.py` they have a chance to run before LittlevGL even starts.

It still doesn't explain [the other issue I saw](https://github.com/littlevgl/lv_binding_micropython/pull/72#issuecomment-623487691), where refreshing stops completely. I haven't been able to reproduce that a second time, though, so I'll just assume it's operator error at the moment. :slightly_smiling_face:

prolomb:
This hypothesis could be verified introducing 1 or 2 seconds delay as first instruction within `example3.py` and run it as `main.py`. In this way I expect to see no issue with preloader. I will give it a try asap

prolomb:
@kisvegabor 
Is there an easy way to patch the code and implement what you did in 7 to fix the issue?
Just to see if it works in our case too.

TA1DB:
> [firmware.hex.zip](https://github.com/littlevgl/lv_binding_micropython/files/4580578/firmware.hex.zip)

Flashed the code you've shared above, loaded  the example3 as "import example3" from a SDCard , system is running for more than a few hours without any problem. I can make further tests if required. https://drive.google.com/open?id=1Pc5G4E08zcoFuppE9WfAxeI1RLxjLfTfand

embeddedt:
I believe this is the change @kisvegabor is referring to: https://github.com/littlevgl/lvgl/commit/5ad1cb76c2d7a0cc0d17138589bc0833208aa883

kisvegabor:
> I believe this is the change @kisvegabor is referring to: littlevgl/lvgl@5ad1cb7

Yes, that's it.

> It should be applicable to 6.1 as well.

I agree.

prolomb:
> This hypothesis could be verified introducing 1 or 2 seconds delay as first instruction within example3.py and run it as main.py. In this way I expect to see no issue with preloader. I will give it a try asap

Even with 10 seconds delay the issue still persist! 
However if I insert 1s delay just before preloader creation then it works! o.O

embeddedt:
Have you tried it with @kisvegabor's fix? If so, this is going to be a hard bug to track down.

prolomb:
> Have you tried it with @kisvegabor's fix? 
Not yet!

prolomb:
Implemented the patch but the problem still persist!

amirgon:
@prolomb - 
Could the problem be related to slow external RAM for display buffer?  
Or does the same problem happen when you use SRAM instead ?

embeddedt:
@amirgon The user interface remains snappy even when the animation isn't running, and flipping pages seems to happen instantly. If the performance of external RAM was the problem, I would expect the UI to be very laggy. It's worth a try, though.

prolomb:
@amirgon 
> Could the problem be related to slow external RAM for display buffer?

I don't think so because the issue in that case should appear whether you run `main.py` or not

> Or does the same problem happen when you use SRAM instead ?

However, I just tried with `malloc` and the issue remain.

prolomb:
@amirgon 
I was starting the new PR but run into a problem.
When i press fork button in https://github.com/littlevgl/lv_micropython this seems to have no effect because i have already a micropython fork (the one i used to test the driver).
May you help me? Is there a guide I can refer to?
Thank you

embeddedt:
@prolomb You can just create a new branch on your existing fork, and then send the PR using that branch.

prolomb:
@embeddedt 

> @prolomb You can just create a new branch on your existing fork, and then send the PR using that branch.

Sorry, this is not clear to me. I thought to add changes respect to https://github.com/littlevgl/lv_micropython

May you type the sequence of steps in detail?
Thank you for the support

embeddedt:
I see the problem. Your fork points to upstream MicroPython and not the LittlevGL version.

If the only change you made on your current fork was to test your binding changes, the simplest solution is probably to just delete that fork and create a new one from `lv_micropython`.

prolomb:
@embeddedt 
I tried with lv_micropython and the freezing of the preloader on startup seems disapperead.
However if i digit on the keyboard the preloader stop spinning. 

@TA1DB
Thank you for the debug. Have you the chance to test the other examples?

embeddedt:
> However if i digit on the keyboard the preloader stop spinning.

I'll have to step through with a debugger and figure out why it doesn't spin, as this will probably cause random breakage in other applications too.

TA1DB:
> @TA1DB
> Thank you for the debug. Have you the chance to test the other examples?
example2 can't import loadpng, the rest works excellent.

amirgon:
@prolomb @kisvegabor @embeddedt - 
Is it ready to be merged? Is there anything missing?
I think it's better to merge this before merging v7.

embeddedt:
@amirgon We can merge it, but the bug with apps randomly freezing is still present. That would need to be tracked down before this is useful. I'd suggest adding a disclaimer that the STM32F7DISC port is still experimental and does not work fully yet.

prolomb:
> We can merge it, but the bug with apps randomly freezing is still present. 

Actually the "random freezing" was fixed (root cause was garbage collection).

Animation of example3.py may not start if it is  loaded as `main.py`, thus I agree with @embeddedt  to add the disclaimer and merge. @amirgon may you add this your own?

@kisvegabor Do I need to sign CLA?

kisvegabor:
> @kisvegabor Do I need to sign CLA?

Yes, please.

amirgon:
I've merge this.
Thank you!

TA1DB:
Hi
Thanks to everybody contributing this collective work.

I got below shown errors in linking stage. Is this a known issue? Ubuntu 16.04LTS.
`
CC build-STM32F7DISC/frozen_content.c
LINK build-STM32F7DISC/firmware.elf
arm-none-eabi-ld: build-STM32F7DISC/stm32_it.o: in function `DMA2D_IRQHandler':
stm32_it.c:(.text.DMA2D_IRQHandler+0x8): undefined reference to `hdma2d'
arm-none-eabi-ld: build-STM32F7DISC/stm32_it.o: in function `LTDC_IRQHandler':
stm32_it.c:(.text.LTDC_IRQHandler+0x8): undefined reference to `hltdc'
arm-none-eabi-ld: build-STM32F7DISC/lib/lv_bindings/lvgl/src/lv_widgets/lv_chart.o: in function `get_next_axis_label':
lv_chart.c:(.text.get_next_axis_label+0x50): undefined reference to `strncpy'
Makefile:628: recipe for target 'build-STM32F7DISC/firmware.elf' failed
make: *** [build-STM32F7DISC/firmware.elf] Error 1

`

prolomb:
@TA1DB 
You are right there is an issue.
Please see https://github.com/lvgl/lv_binding_micropython/issues/82

TA1DB:
> @TA1DB
> You are right there is an issue.
> Please see #82

Thanks, I hope it will be resolved soon because I look forward to being able to implementg lv_micropython on my STM32F7DISC with my own build.

PR #76, opened 2020-05-18
Fix usleep call in modfb

darchons:
The unit for `usleep` is microseconds, so the argument should be `1000` to sleep for 1 millisecond.


embeddedt:
Looks correct to me. Thanks!

PR #79, opened 2020-06-20
Add MP_ERROR_TEXT for MicroPython post-1.12 (esp32-only)

tve:
This PR has fixes needed post MicroPython v1.12 due to the required MP_ERROR_TEXT macro.
This PR only fixes the esp32 port, same changes need to happen for other ports as well..
I realize you don't want to just merge this given that you're most likely still working with v1.12, I'm just providing it for the future.

### DIFF
--- a/driver/esp32/modILI9341.c
+++ b/driver/esp32/modILI9341.c
@@ -184,12 +184,12 @@ STATIC void disp_spi_init(ILI9341_t *self)
 	//Initialize the SPI bus
 	ret=spi_bus_initialize(self->spihost, &buscfg, 1);
     if (ret != ESP_OK) nlr_raise(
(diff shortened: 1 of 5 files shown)


amirgon:
Thank you.

In the meanwhile I'll add a define for `MP_ERROR_TEXT` in `lv_conf.h`
That way I'll be able to merge your change now, and in the future just remove that define.

embeddedt:
This is causing build breakage on the JavaScript port (after 1.12). `lv_conf.h` is included before the internal MicroPython headers, so it defines `MP_ERROR_TEXT`. When the MP header gets included, it tries to redefine the macro and causes errors.

At the moment I'm working around this by commenting out the definition in `lv_conf.h`.

embeddedt:
Ignore my previous message; it turns out that I forgot to update submodules.

PR #90, opened 2020-08-13
Support for ili9488 display driver.

mhepp63:
Reworked sample ili9341.py driver for ili9341 display drivers to support ili9341 and ili9488 display chips.

Difference between ili9341 and ili9488 is minimal:

  * ili9488 has another init sequence
  * ili9488 do not support RGB565 (16bits) colors in SPI mode, only RGB666. Data transfer in while bytes, so data transfered to display has format RGB888.
    * details are in ili9488 datasheet.
    * then ili9488 driver require 32bits color support
  * ili9488 has own ulsh function - for stripping alpha byte from each pixel
    * converts each pixel from RGBA to RGB before sending data.
  * Actual ili9488 driver supports only hybrid mode.

Usage:

import ili9XXX

... and with ili9488 display

display = ili9XXX.ili9488(
        spihost=esp.HSPI_HOST, mhz=40, factor=8, hybrid=True, width=320, height=480,
        colormode=COLOR_MODE_RGB, rot=PORTRAIT, invert=False, double_buffer=True)

... with ili9341 display

display = ili9XXX.ili9341(
        spihost=esp.HSPI_HOST, mhz=40, factor=8, hybrid=True, width=240, height=320,
        colormode=COLOR_MODE_RGB, rot=PORTRAIT, invert=False, double_buffer=True)

(used values in examples are default)


driver/esp32/espidf.c
```diff
@@ -203,3 +203,63 @@ void ili9341_flush(void *_disp_drv, const void *_area, void *_color_p)
     ili9341_send_data_dma(disp_drv, color_p, size * 2, dc, *spi_ptr);
 }
 
+
+typedef struct {
+	uint8_t blue;
+	uint8_t green;
+	uint8_t red;
+} lv_color24_t;
```
embeddedt, line 211: Potential conflict with LVGL if the [`feat/rgb888`](https://github.com/lvgl/lvgl/issues/1722) branch gets merged. I'd suggest just calling this `color24_t` and leave the `lv_` prefix off.

mhepp63: OK, no problem...

amirgon:
Hi @mhepp63 , thank you for this contribution!

Question -   

If I understand correctly, you assume 32bit color mode.
ili9341 requires setting `LV_CFLAGS="-DLV_COLOR_DEPTH=16 -DLV_COLOR_16_SWAP=1"` when building Micropython.  
How did you set these compilation flags (or others) for ili9488 when building Micropython? Or did you leave the default values?


I have a few comments:

- `ili9488_flush` is almost identical to `ili9341_flush`.  
The only difference is the additional ARGB to RGB conversion you've added.  
To avoid code duplication, please merge these flush functions to a single flush function.  
You can pass another flag on `disp_drv.user_data` the [same way `dc` and `spi` are passed](https://github.com/lvgl/lv_binding_micropython/blob/7f7af8ee49d7e932f5ac5d6643c900376ff568f0/driver/esp32/ili9341.py#L133), extract it [on the C code](https://github.com/lvgl/lv_binding_micropython/blob/7f7af8ee49d7e932f5ac5d6643c900376ff568f0/driver/esp32/espidf.c#L170-L176) and condition the ARGB to RGB conversion by this flag in the flush function.  
This would give us a single flush function which can be configured for different displays.

- Could you add some text to the comment at the beginning of `ili9XXX.py`? Currently it is a copy of the comment on `ili9341.py` and I think it's worth adding a few words about ili9488, color mode requirements etc.

- I would like to add your change to current version (and not wait for another major release), but for this we must make sure it's backward compatible and doesn't break existing code.  
Current code [uses something like `from ili9341 import ili9341`](https://github.com/lvgl/lv_binding_micropython/blob/7f7af8ee49d7e932f5ac5d6643c900376ff568f0/examples/advanced_demo.py#L303).  
Now we would need `from ili9XXX import ili9341`, which is nicer and makes more sense, but we still need to support the old form of `from ili9341 import ili9341` at least for current version.  
I would like to avoid duplicated code and remove the old ili9341.py, but we would still need some stub instead of it to make `from ili9341 import ili9341` delegate to ili9XXX.ili9341.

mhepp63:
Hi @amirgon, thank you...

And your questions... Yes, my module uses 32bit color mode. All other settings can be default. It is sad, but it is not possible to use same build with ili9341 and ili9488. I made tests with fresh source, build with default settings. Only thing I checked/changed was color depth. Performance is not for huge animantions, refresh rate is cca 7-8fps (with esp32, not exactly measured, theoretical maximum is 10fps@40MHz SPI).

- _ili9488_flush is almost identical to ili9341_flush..._ yes, I uderstand. And in case of megre [feat/rgb888],(https://github.com/lvgl/lvgl/issues/1722) it will be exactly the same. OK, no problem, I will do it.

- _Could you add some text to the comment... _ usage/requirements/compilation parameters/etc... OK, no problem.

- _I would like to add your change to current version..._  ...cool! I did not expect this ;) You mean something for backward compatibility... This is why I left old ili9341.py untouched, but if you want to replace them, I will find some solution.

amirgon:
> Performance is not for huge animantions, refresh rate is cca 7-8fps (with esp32, not exactly measured, theoretical maximum is 10fps@40MHz SPI).

- ESP32 is actually capable of up to 80Mhz SPI output frequency [with the right configuration](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/spi_master.html#dummy-bit-workaround:~:text=If%20the%20Host%20only%20writes%20data%2C,if%20the%20GPIO%20matrix%20is%20used.).
- lvgl tries to update only what is necessary, not the entire screen. So small animations (such as button click for example) could still be fast enough.  
Full page scrolling, on the other hand, would probably be too slow.

> * And in case of megre [feat/rgb888](https://github.com/lvgl/lvgl/issues/1722), it will be exactly the same

After https://github.com/lvgl/lvgl/issues/1722 is merged, it would be a good idea to remove the ARGB to RGB conversion and save some cycles.

mhepp63:
Ok, new commits: 
 
 - Single flush function ili9xxx_flush() with color convert for ili9488.
 - Added description and usage to ili9XXX.py header (I left 40MHz as default value, my display do not work properly at 80MHz).
 - ili9341.py is using new library now. Lightly tested, so I hope, it will not break compatibility.
 - ...and typo fix as a bonus ;)

amirgon:
Merged.
Thank you!

PR #114, opened 2020-12-17
split gui initialization into 2 files to allow change of width, heigh…

uraich:
split GUI initialization into 2 files and added t-watch 2020 initialization.
Default driver parameters are used. The init parameters for the Lolin TFT 2.4 are given in comments.
Default width and height parameters for SDL are set to 480x320 as requested.
I tested this for SDL, t-watch 2020 and Lolin TFT 2.4 ili9341 display + xpt2046. For landscape mode I added my calibration parameters for portrait mode I rely on the default ones.


amirgon:
I've added some small fixes on 4ec889c.

Merging.

uraich:
Yes, raising a runtime error is the better solution.
Thank you for your help

PR #126, opened 2021-02-22
update modSDL.c to support Microsoft Windows

retsyo:
now I can, at least in MSYS2+Mingw64, compile and get `micropython.exe` with LVGL. Then I can learn and test pyLVGL on windows

### DIFF
--- a/driver/SDL/modSDL.c
+++ b/driver/SDL/modSDL.c
@@ -3,7 +3,9 @@
 #define _POSIX_C_SOURCE 200809L
 #endif
 #include <errno.h>
+#ifndef _WIN32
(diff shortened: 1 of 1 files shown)


amirgon:
Hi @retsyo 

The purpose of the code you excluded is to allow LVGL event loop to work while the REPL is waiting for input.  
Without it, the event loop would freeze whenever the application is waiting for user input.

I understand that Windows does not support posix `sigaction`.
Did you consider alternative approaches for running LVGL event loop while blocked on input?

retsyo:
sorry, it runs out of my knowledge. So let's wait for a real and better solution for windows os.

PR #130, opened 2021-03-15
Make lvgl tick more accurate. (rebase to master)

rockindy:


driver/SDL/modSDL.c
```diff
@@ -40,9 +40,22 @@ STATIC int tick_thread(void * data)
 {
     (void)data;
 
+    Uint64 pfreq = SDL_GetPerformanceFrequency();
+    Uint64 tick_ms = pfreq / 1000 * LV_TICK_RATE;
```
amirgon, line 44: Just for clarity I would add parenthesis.

amirgon:
In general looks good to me.  
What do you think @embeddedt ?

embeddedt:
I am still not convinced about the `* 2` logic. Why can't you just call `lv_tick_inc(acc)` and get rid of the accumulated ticks? LVGL can handle larger-than-normal values being passed to `lv_tick_inc`.

rockindy:
@amirgon , I have added parenthesis, thanks.

@embeddedt , I think avoid using `* 2` logic is great suggestion and has updated new coomit.
`delta - tick_ms` is very small error which is less than 1ms, the `acc` needs several ticks to reach 1ms and more ticks to reach `LV_TICK_RATE`. We should only call `lv_tick_inc()` when the unit of parameter is ms.

amirgon:
@embeddedt - Do you have any more comments about this? Can I merge this?

embeddedt:
Looks good.

amirgon:
Merged.
Thank you!

PR #131, opened 2021-04-11
fix(stm32) skip lv_task_handler if scheduler queue is full

### LINKED ISSUE #82: Current STM32 port issues
These were reproduced on STM32F7DISCO, but may also apply to other boards.

* `advanced_demo` crashes if run after a soft reset (the board must be hard reset). I suspect that the display driver isn't deinitializing/reinitializing something correctly.
* Running the following code after `advanced_demo` causes the entire UI to freeze and stop responding to input (though the REPL continues working):

```py
import lvgl as lv
obj = lv.obj()
lv.scr_load(obj)
btn = lv.btn(obj)
```

Entire UI freezes also happened with v6, along with animations ceasing to run.

I don't think we checked whether `advanced_demo` worked with a soft reset on v6. This isn't that major because a hard reset only takes about a second on these STM32 boards, so it's more of an inconvenience.

embeddedt:
This means that GUI ticks may be skipped on a busy system, but that is better than hanging the GUI forever.

Fixes #82.

### DIFF
--- a/driver/stm32/lvstm32.py
+++ b/driver/stm32/lvstm32.py
@@ -21,7 +21,9 @@ class lvstm32():
     def timer_cb(self, t):
         lv.tick_inc(self.delay)
         # Passing self.task_handler would cause allocation.
-        micropython.schedule(self.task_handler_ref, 0)
+        try:
+            micropython.schedule(self.task_handler_ref, 0)
+        except RuntimeError:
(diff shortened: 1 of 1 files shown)


amirgon (approved):

amirgon:
That's great!
Merging.

PR #132, opened 2021-04-17
Adding driver for Zephyr OS

svetelna:
This PR introduces a driver for Zephyr OS so that Micropython+LVGL could be used as a Zephyr application.
More information is in the enclosed README.

The lvgl.c file is taken from the Zephyr codebase and adjusted for this use case.

The PR was tested on i.MX RT1050, with ELCDIF display driver and FT5336 input driver from Zephyr OS.

Edit:
Zephyr already supports LVGL (in C language) - driver abstraction layer is available in Zephyr codebase and LVGL as a Zephyr's module can be enabled with Kconfig.

For Zephyr + Micropython + LVGL use case are LVGL source files taken from the LV bindings project; thus LVGL has to be disabled in Zephyr's Kconfig configuration in order to prevent redefinitions. The reason is, that LVGL should in this use case use Micorpython's memory functions and not Zephyr's, so it seemed more appropriate to preserve LVGL + Micropython and not LVGL + Zephyr connection.

### DIFF
--- a/driver/zephyr/lvgl.c
+++ b/driver/zephyr/lvgl.c
@@ -2,6 +2,14 @@
  * Copyright (c) 2018-2019 Jan Van Winkel <jan.van_winkel@dxplore.eu>
  *
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * changes:
+ *		Zephyr's logging -> printf
+ *		CONFIG_LVGL_HOR_RES_MAX -> LV_HOR_RES_MAX
+ *		CONFIG_LVGL_VER_RES_MAX -> LV_VER_RES_MAX
+ *		added debug log
+ *		removed SYS_INIT for lvgl_init
+ *		lvgl_init no longer static
  */
 
 #include <init.h>
@@ -14,14 +22,11 @@
 #ifdef CONFIG_LVGL_USE_FILESYSTEM
 #include "lvgl_fs.h"
 #endif
+#define CONFIG_LVGL_POINTER_KSCAN 1
 #ifdef CONFIG_LVGL_POINTER_KSCAN
 #include <drivers/kscan.h>
 #endif
 #include LV_MEM_CUSTOM_INCLUDE
-
-#define LOG_LEVEL CONFIG_LVGL_LOG_LEVEL
-#include <logging/log.h>
-LOG_MODULE_REGISTER(lvgl);
 
 #ifdef CONFIG_LVGL_BUFFER_ALLOC_STATIC
 
@@ -42,36 +47,6 @@ static uint8_t buf1[BUFFER_SIZE] __aligned(4);
 #endif
 
 #endif /* CONFIG_LVGL_BUFFER_ALLOC_STATIC */
-
-#if CONFIG_LVGL_LOG_LEVEL != 0
-static void lvgl_log(lv_log_level_t level, const char *file, uint32_t line,
-		const char *func, const char *dsc)
-{
-	/* Convert LVGL log level to Zephyr log level
-	 *
-	 * LVGL log level mapping:
-	 * * LV_LOG_LEVEL_TRACE 0
-	 * * LV_LOG_LEVEL_INFO  1
-	 * * LV_LOG_LEVEL_WARN  2
-	 * * LV_LOG_LEVEL_ERROR 3
-	 * * LV_LOG_LEVEL_NUM  4
-	 *
-	 * Zephyr log level mapping:
-	 * *  LOG_LEVEL_NONE 0
-	 * * LOG_LEVEL_ERR 1
-	 * * LOG_LEVEL_WRN 2
-	 * * LOG_LEVEL_INF 3
-	 * * LOG_LEVEL_DBG 4
-	 */
-	uint8_t zephyr_level = LOG_LEVEL_DBG - level;
-
-	ARG_UNUSED(file);
-	ARG_UNUSED(line);
-	ARG_UNUSED(func);
-
-	Z_LOG(zephyr_level, "%s", log_strdup(dsc));
-}
-#endif
 
 #ifdef CONFIG_LVGL_BUFFER_ALLOC_STATIC
 
@@ -298,8 +273,10 @@ static int lvgl_pointer_kscan_init(void)
 	}
 
 	int err = kscan_enable_callback(kscan_dev);
-	if (err != 0)
-		printf("callback enabling failed\n");
+	if (err != 0) {
+		printf("Callback enabling failed\n");
+		return err;
+	}
 
 	return 0;
 }
--- a/driver/zephyr/lvgl.c
+++ b/driver/zephyr/lvgl.c
@@ -1,0 +1,356 @@
+/*
+ * Copyright (c) 2018-2019 Jan Van Winkel <jan.van_winkel@dxplore.eu>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <init.h>
+#include <stdio.h>
+#include <zephyr.h>
+#include <lvgl.h>
+#include "lvgl_display.h"
+#include "lvgl_driver.h"
+#include <drivers/display.h>
+#ifdef CONFIG_LVGL_USE_FILESYSTEM
+#include "lvgl_fs.h"
+#endif
+#ifdef CONFIG_LVGL_POINTER_KSCAN
+#include <drivers/kscan.h>
+#endif
+#include LV_MEM_CUSTOM_INCLUDE
+
+#define LOG_LEVEL CONFIG_LVGL_LOG_LEVEL
+#include <logging/log.h>
+LOG_MODULE_REGISTER(lvgl);
+
+#ifdef CONFIG_LVGL_BUFFER_ALLOC_STATIC
+
+lv_disp_buf_t disp_buf;
+
+#define BUFFER_SIZE (CONFIG_LVGL_BITS_PER_PIXEL * ((CONFIG_LVGL_VDB_SIZE * \
+			LV_HOR_RES_MAX * LV_VER_RES_MAX) / 100) / 8)
+
+#define NBR_PIXELS_IN_BUFFER (BUFFER_SIZE * 8 / CONFIG_LVGL_BITS_PER_PIXEL)
+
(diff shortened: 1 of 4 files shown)


driver/zephyr/lvgl.c
```diff
@@ -0,0 +1,356 @@
+/*
+ * Copyright (c) 2018-2019 Jan Van Winkel <jan.van_winkel@dxplore.eu>
+ *
+ * SPDX-License-Identifier: Apache-2.0
```
embeddedt, line 4: @amirgon Just wanted to bring this to your attention: are you okay with having an Apache-licensed file in `lv_binding_micropython`?

amirgon: Thanks for this notice @embeddedt 

Both Apache-2.0 and MIT are permissive licenses.
Although not identical, I believe an MIT project can have Apache-2.0 sources as long as the [Apache-2.0 requirements](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)) are met:
- The copyright notice appears on the copyrighted files
- Any significant changes made to the copyrighted files must be stated
- Notice file

@svetelna - Were any changes made to the copyrighted files? Is there a NOTICE file on the original repo where you copied the copyrighted files from?

amirgon:
Thank you for this contribution @svetelna !

With Micropython we try to move configuration to runtime as much as possible.  
The idea is to have a single firmware image that could be used with multiple types of display. The selection of the display type and display parameters (such as SPI pins for example, landscape/portrait orientation etc.) is usually determined on runtime by Python code.

As an example, see [ili9xxx driver for ESP32](https://github.com/lvgl/lv_binding_micropython/blob/master/driver/esp32/ili9XXX.py).  
The user constructs a Python display driver on runtime (either `ili9488` or `ili9341`) and passes it [many configuration parameters](https://github.com/lvgl/lv_binding_micropython/blob/216b3c509606e58836f95fd725827a4a66f89444/driver/esp32/ili9XXX.py#L71-L76). 

How does this align with Micropython+Zephyr?  
I noticed that you rely on many compile time macros (for example the device name "FT5336") instead of receiving them on runtime, and your Micropython module `ZDD` really contains only an "init" function that calls `lvgl_init` with no arguments. So it seems that the user has to configure everything on compile time and rebuild the firmware on every configuration change.

driver/zephyr/lvgl.c
svetelna: Thank you for your response.
I did not find any NOTICE file for Zephyr.
However, I have changed the lvgl.c file a bit:
* Zephyr's LOG statements are changed to printfs
* some error checks are added
* removed ```SYS_INIT(lvgl_init, APPLICATION, CONFIG_APPLICATION_INIT_PRIO)``` macro, so that the LVGL init is not called during boot of the system.

I can state these changes in the lvgl.c.

The main reason I copied the file, and have not used its original location within Zephyr's codebase is Zephyr's Kconfig configuration. The LVGL has to be disabled in Zephyr (in order to use LVGL from the LV bindings project), thus the driver init would not work. I should have stated this in the pull request comment. I will add it there.

svetelna:
@amirgon Thank you for your feedback.
In Zephyr, needed drivers have to be chosen with Kconfig, thus have to be known before compilation. So, the user needs to make sure, the compiled image already contains the drivers, that he needs. In the context of this PR, these have to be stated in Micropython's Zephyr port's ```prj.conf``` file or similar.
I think it would be possible to enable multiple drivers in Zephyr's Kconfig (but still chosen by the user before compilation) and then, as you pointed out, the ZDD.init could take as parameters names of the specific drivers.
If it seems OK with you, I could try to change the implementation.
Also, display resolution could be configurable, with some default values, possibly?

amirgon:
> I think it would be possible to enable multiple drivers in Zephyr's Kconfig (but still chosen by the user before compilation) and then, as you pointed out, the ZDD.init could take as parameters names of the specific drivers.
> If it seems OK with you, I could try to change the implementation.

Yes that sounds good.

> Also, display resolution could be configurable, with some default values, possibly?

Yes. And other parameters.
For example, where does the user specify through which pins and interfaces the display is connected? Other parameters include orientation and color mode, interface parameters such as speed, etc.

driver/zephyr/lvgl.c
amirgon: > I can state these changes in the lvgl.c.

Yes that would be required from legal perspective.



> The main reason I copied the file, and have not used its original location within Zephyr's codebase is Zephyr's Kconfig configuration.

This is not clear to me.
Isn't it possible to change the original lvgl.c such that it could support LVGL with or without Micrpython?  
Duplicating code makes maintanance difficult. Future changes to the original lvgl.c would need to be merged into your copy.

svetelna: Problem with using original lvgl.c is in Kconfig defines used in the file, e.g., CONFIG_LVGL_POINTER_KSCAN.
These can not be enabled or used in Zephyr's build system, if their parent, Kconfig entry LVGL, is disabled.

I see no way how to define these within original lvgl.c. However, when the lvgl.c is copied to the Zephyr driver directory, the entries can be defined in the header file, or values directly from lv_conf.h can be used. 

If the LVGL entry in Kconfig is enabled, then the LVGL sources are compiled within Zephyr - causing redefinitions.

svetelna:
>Yes. And other parameters.
>For example, where does the user specify through which pins and interfaces the display is connected? Other parameters include >orientation and color mode, interface parameters such as speed, etc.

Zephyr uses devicetrees, so the pins and display connections need to be defined before compilation. With this in mind, I am rethinking, if driver names as parameters in lvgl_init are a good idea. Other display driver configurations are being done via Kconfig, also before the compilation.

driver/zephyr/lvgl.c
amirgon: > I see no way how to define these within original lvgl.c. However, when the lvgl.c is copied to the Zephyr driver directory, the entries can be defined in the header file, or values directly from lv_conf.h can be used.

There may be a way around this.
You could add to the original lvgl.c under Zephyr something like this:
```c
#if ENABLE_LV_MICROPYTHON == 1
#include "zephyr_lv_micropython.h"
#endif
```
Where `zephyr_lv_micropython.h` contains the missing definition.  
Your PR requires changes to the Makefile anyway, so it's pretty easy to add to the Makefile something like:
```makefile
CFLAGS_MOD += -DENABLE_LV_MICROPYTHON=1
```

This can prevent duplicating lvgl.c.  
What do you think?

amirgon:
> Zephyr uses devicetrees, so the pins and display connections need to be defined before compilation.

In my opinion this is unfortunate and doesn't align well with the Micropython way of configuring most things on runtime.

> I tested this PR on v1.14, thus with Makefiles. With the upgrade to CMake, only the README would have to be updated. I will update it.

Instead of guiding the user how to edit the Makefile or CMake scripts, why don't you integrate these changes into the [Zephyr port on lv_micropython](https://github.com/lvgl/lv_micropython/tree/master/ports/zephyr)?

In many cases lv_micropython is used directly to run Micropython+LVGL, and on other cases it could at least be a good reference.

driver/zephyr/lvgl.c
svetelna: I think that the Zephyr itself should not have any links or references to the LV binding project, as they are not connected directly but only via the Micropython. Even Micropython is currently not mentioned anywhere in the Zephyr codebase.
It seems that from the start, the goal was to make the connection from the opposite side -> that the Micropython would contain references to Zephyr, built as the Zephyr's application.
Therefore I cannot entirely agree with the proposed solution.

embeddedt:
MicroPython 1.15 will be used starting with LVGL 8; a beta version of the bindings is in `dev-8.0`.

svetelna:
Thanks for pointing me out @embeddedt.

svetelna:
Hello, I have just created a PR for the lv_micropython project: https://github.com/lvgl/lv_micropython/pull/25
Here the lv_bindings are integrated into Micropython v1.14.

I have also tried to integrate changes into Micropython 1.15 (with CMake build); however, this attempt has failed on the fact that the current Zephyr stable version (2.5) does not support LVGL 8. Thus the bindings between Zephyr display drivers and LVGL flushing and reading functions are invalid.
Otherwise, I think that the solution works, and I can create a new pull request for the lv_micropython dev-8.0 branch, which can be merged later (when the LVGL 8 is supported by the Zephyr). Just let me know what would be OK for you.

amirgon:
> Otherwise, I think that the solution works, and I can create a new pull request for the lv_micropython dev-8.0 branch, which can be merged later (when the LVGL 8 is supported by the Zephyr). Just let me know what would be OK for you.

@svetelna  LVGL v8 wasn't released yet, so it might be too early for LVGL v8 integration in Zephyr.  
When Zephyr supports LVGL 8 feel free to open a new PR. By then we would probably already merge dev-8 branch into main on lv_micropython and lv_micropython_binding.

On the README you specify changes than need to be done in the Zephyr Micropython port. It's worth mentioning there that this is already implemented in lv_micropython (once we merge the lv_micropython PR), so the user is not required to do all that when using lv_micropython.  
Is there anything the user still needs to do if he uses lv_micropython?

@embeddedt - Ok to merge this? or do you want more time to review this?

embeddedt:
I think merging this is fine. It sounds like MIT and Apache are compatible enough, and also no one is likely to use that file outside of Zephyr anyway.

svetelna:
@amirgon  The user needs set properly the display and input driver entries in lvgl_driver.h (in lv_bindings/drivers/zephyr) and in specific-desk.prj (in micropython/ports/zephyr). Moreover, the lv_conf.h needs to be edited, to user Zephyrs LV_TICK_CUSTOM_INCLUDE.

I have already updated the readme file with information about lv_micropython project.

amirgon:
Merged! Thank you for your contribution @svetelna!  

Please consider adding a CI job that checks that Zephyr+Micropython+LVGL builds correctly.  
Today we have such test for the unix port, and we are planning to add ESP32 / STM32 tests.  

Personally I don't use Zephyr so I would not be able to tell when Zephyr port breaks without an automated test.

svetelna:
Thanks for merging.
I will try to look into the CI checks for the Zephyr port.

PR #137, opened 2021-04-23
fix(stm32) changes for v8

embeddedt:
With these changes merged and applied to `lv_micropython` the STM32 port now runs without problems.

This PR:

* Removes GPU related callbacks as they are now redundant with LVGL's GPU support. Support for LVGL GPUs will be added back soon.
* Switches buffer sizes as full double buffering behavior has changed in v8.

(cc @prolomb)

### DIFF
--- a/examples/advanced_demo.py
+++ b/examples/advanced_demo.py
@@ -365,15 +365,13 @@ class AdvancedDemoApplication:
         self.tick = lvstm32.lvstm32()
         lcd.init(w=hres, h=vres)
         disp_buf1 = lv.disp_draw_buf_t()
-        buf1_1 = lcd.framebuffer(1)
-        buf1_2 = lcd.framebuffer(2)
+        buf1_1 = bytearray(hres * 50 * lv.color_t.SIZE)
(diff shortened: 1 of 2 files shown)


examples/advanced_demo.py
```diff
@@ -365,15 +365,13 @@ def init_gui_stm32(self):
         self.tick = lvstm32.lvstm32()
         lcd.init(w=hres, h=vres)
         disp_buf1 = lv.disp_draw_buf_t()
-        buf1_1 = lcd.framebuffer(1)
-        buf1_2 = lcd.framebuffer(2)
+        buf1_1 = bytearray(hres * 50 * lv.color_t.SIZE)
```
amirgon, line 368: Why 50?

embeddedt: RAM is not an issue on this board so I would normally have used a full buffer size, but my profiling from v7 development showed that buffer sizes of 1/4 to 1/5 barely decrease the rendering speed, so I picked `hres * 50` which is approximately 1/5 of the total display size.

PR #139, opened 2021-05-22
Replace non-working indev example by a simpler working one.

ropg:
The example at `driver/generic/lvindev_example.py` does not work. It uses a python module that is not registered by default to do things that do not work with the present lvgl python module. This sends people the wrong way, also because the present way of doing things is not properly documented to my knowledge. I deleted the example and replaced it by the bare minimum for a working driver.

Please feel free to modify if you feel things need to be done differently, I am new to this codebase.

See [this forum post](https://forum.lvgl.io/t/missing-indev-drv-init/5537/3) for further details.


amirgon:
Hi @ropg !
Thank you for this contribution!

Here is another example of pure python indev driver: https://github.com/lvgl/lv_binding_micropython/blob/master/driver/linux/evdev.py

driver/generic/indev_example.py
```diff
@@ -0,0 +1,34 @@
...
+#       from my_indev import my_indev
+#       touch = my_indev()
+
+
+import lvgl as lv
+
+class my_indev:
+
+    def __init__(self):
+        indev_drv = lv.indev_drv_t()
```
amirgon, line 22: I think it's important to keep a reference to `indev_drv_t` and prevent gc from collecting it.  
The simplest way it to keep it as a member of `self`, as done here:

https://github.com/lvgl/lv_binding_micropython/blob/dadad31bdce92fc96ffac6b2805ac0d5d4b884eb/driver/linux/evdev.py#L46

driver/generic/indev_example.py
```diff
@@ -0,0 +1,34 @@
...
+        indev_drv.init()
+        indev_drv.type = lv.INDEV_TYPE.POINTER
+        indev_drv.read_cb = self.callback
+        indev_drv.register()
+
+    def callback(self, driverptr, dataptr):
+        # This is where you need to get the actual position from
+        # the hardware. This just simulates pressed at 50, 100.
+        data = lv.indev_data_t.cast(dataptr)
```
amirgon, line 31: Are you sure casting is needed?

Here `data` is accessed directly:

https://github.com/lvgl/lv_binding_micropython/blob/dadad31bdce92fc96ffac6b2805ac0d5d4b884eb/driver/linux/evdev.py#L52

amirgon (changes requested):

driver/generic/indev_example.py
ropg: done

driver/generic/indev_example.py
ropg: indeed, works without

ropg:
OK, all fixed, ready to be merged, I think.

PR #140, opened 2021-05-22
Add pure Python FT6X36 touch IC (i2c) LVGL indev driver

ropg:
Here's a driver for the FocalTech FT6X36 capacitive touch ICs, which communicate via i2c. The driver itself is documented in a comment at the top of the code as follows:

```text
# Pure Python LVGL indev driver for the FocalTech FT6X36 capacitive touch IC
#
#       from ft6x36 import ft6x36 
# 
#       touch = ft6x36(sda=<pin>, scl=<pin>)
#
# If you set the size of the touchpad, you have the option to invert each 
# axis, and you'll get some extra robustness against occasional read errors
# as values outside the touchpad are quietly rejected. If you select to swap
# the axes, width and height as well as the inversions refer to the situation
# before the swap.
#
# The nice thing about this driver is that it allows access to the second
# finger, as the FT6X36 is multi-touch. (Two fingers max, with caveats on
# some boards.)
#
# The number of presses is in touch.presses, touch.points[0] and points[1]
# hold the positions. LVGL is not (yet) multi-touch, so all it sees is the
# position in points[0].
```

Also included are minor changes to the README to anounce the availability and give a small example, in this case for an M5Stack Core2 device, which uses this chip:

> Here's another example, this time importing and initialising display and touch drivers for the M5Stack Core2 device, which uses an FT6336 chip on the I2C bus to read from its capacitive touch screen and uses an ili9342 display controller, which has some inverted signals compared to the ili9341:

```python
from ili9XXX import ili9341
disp = ili9341(mosi=23, miso=38, clk=18, dc=15, cs=5, invert=True, rot=0x10)

from ft6x36 import ft6x36
touch = ft6x36(sda=21, scl=22, width=320, height=280)
``` 

The driver was tested as submitted and works on my M5Core2 device.  Please feel free to edit or tell me what to do differently if you want things in another way: I am new to this codebase.


amirgon:
Hi @ropg, thank you for your contribution!

You placed the driver under `driver/esp32`. In what way is it ESP32 specific?
I think that the modules `I2C`, `Pin` are also available on other platforms.
If it's not esp32 specific, I think a better location would be `driver/generic`.

@kisvegabor - Do you have any future plans to support multi-touch in LVGL? That might require some changes in the indev API.

ropg:
I placed it there simply because all the other things were there. Other hardware also has SPI, would that not also make the ili9XXX go into generic?

I need to change the other thing about indev_drv needing to be kept in focus and not needing the cast. But I'm happy to place it wherever you think it goes. Shall I place it in generic instead?

kisvegabor:
> Do you have any future plans to support multi-touch in LVGL? That might require some changes in the indev API.

It's not on the roadmap now (and no plan for it).

amirgon:
> I placed it there simply because all the other things were there. Other hardware also has SPI, would that not also make the ili9XXX go into generic?

"driver/esp32" is supposed to contain only ESP32-specific drivers.  
Other hardware under "driver/esp32" import `espidf` module which is esp32 specific and uses the SPI from directly from ESP-IDF API because the Micropython generic SPI was too limited.  

But a better approach is to create a generic driver that can work on multiple hardware platforms, when this is possible. I think FT6X36 belongs to that category so I think it's best to place it there.

> need to change the other thing about indev_drv needing to be kept in focus and not needing the cast.

Yes the few comments on https://github.com/lvgl/lv_binding_micropython/pull/139 apply here as well.

amirgon:
Merged!

Thanks.

PR #143, opened 2021-05-27
Add in support for gc9a01 displays

minyiky:
I have added a driver for gc9a01 displays as a child of ili9XXX as it works in a very similar manner.

Two caveats, I have not been able to delve in to the c code and so hybrid mode is not currently supported. Also my display does not seem to match the datasheet for write direction and colour mode (both are reversed) and so I have written it this way in the driver but it may need to be updated if my display is an odd case

### DIFF
--- a/driver/esp32/ili9XXX.py
+++ b/driver/esp32/ili9XXX.py
@@ -242,7 +242,7 @@ class ili9XXX:
             #     micropython.schedule(post_isr, None)
             # except RuntimeError:
             #     pass
-        
+
         self.spi_callbacks = esp.spi_transaction_set_cb(None, flush_isr)
 
     #
@@ -308,13 +308,13 @@ class ili9XXX:
         self.trans.tx_buffer = data         # data should be allocated as DMA-able memory
         self.trans.user = None
         esp.spi_device_polling_transmit(self.spi, self.trans)
-    
+
     def spi_send_dma(self, data):
         self.trans.length = len(data) * 8   # Length is in bytes, transaction length is in bits. 
         self.trans.tx_buffer = data         # data should be allocated as DMA-able memory
         self.trans.user = self.spi_callbacks
         esp.spi_device_queue_trans(self.spi, self.trans, -1)
-    
+
     ######################################################
     ######################################################
 
@@ -550,7 +550,7 @@ class ili9488(ili9XXX):
     def __init__(self,
         miso=5, mosi=18, clk=19, cs=13, dc=12, rst=4, power=14, backlight=15, backlight_on=0, power_on=0,
         spihost=esp.HSPI_HOST, mhz=40, factor=8, hybrid=True, width=320, height=480,
-        colormode=COLOR_MODE_RGB, rot=180, invert=False, double_buffer=True, half_duplex=True,
+        colormode=COLOR_MODE_RGB, rot=PORTRAIT, invert=False, double_buffer=True, half_duplex=True,
         asynchronous=False, initialize=True
     ):
 
@@ -596,7 +596,7 @@ class gc9a01(ili9XXX):
 
     def __init__(self,
         miso=5, mosi=18, clk=19, cs=13, dc=12, rst=4, power=14, backlight=15, backlight_on=0, power_on=0,
-        spihost=esp.HSPI_HOST, mhz=60, factor=4, hybrid=False, width=240, height=240,
+        spihost=esp.HSPI_HOST, mhz=60, factor=4, hybrid=True, width=240, height=240,
         colormode=COLOR_MODE_RGB, rot=PORTRAIT, invert=False, double_buffer=True, half_duplex=True,
         asynchronous=False, initialize=True
     ):
--- a/driver/esp32/ili9XXX.py
+++ b/driver/esp32/ili9XXX.py
@@ -38,7 +38,8 @@
 # Pure Micropython could be viable when ESP32 supports Viper code emitter.
 #
 # ili9488 driver DO NOT support pure micropython now (because of required
-# color convert). Pure micropython is supported only for ili9341 display!
+# color convert). Pure micropython is only supported the for ili9341 and
+# gc9a01 displays!
 ##############################################################################
 
 import espidf as esp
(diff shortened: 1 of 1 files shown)


driver/esp32/ili9XXX.py
```diff
@@ -537,7 +549,7 @@ class ili9488(ili9XXX):
     def __init__(self,
         miso=5, mosi=18, clk=19, cs=13, dc=12, rst=4, power=14, backlight=15, backlight_on=0, power_on=0,
         spihost=esp.HSPI_HOST, mhz=40, factor=8, hybrid=True, width=320, height=480,
-        colormode=COLOR_MODE_RGB, rot=PORTRAIT, invert=False, double_buffer=True, half_duplex=True,
+        colormode=COLOR_MODE_RGB, rot=180, invert=False, double_buffer=True, half_duplex=True,
```
amirgon, line 552: `PORTRAIT` is currently defined to `MADCTL_MX` which is `const(0x40)`.  
Why did you change it to 180?

driver/esp32/ili9XXX.py
```diff
@@ -576,3 +588,104 @@ def __init__(self,
             spihost, mhz, factor, hybrid, width, height, colormode, rot, invert, double_buffer, half_duplex, display_type=DISPLAY_TYPE_ILI9488,
             asynchronous=asynchronous, initialize=initialize)
 
+
+class gc9a01(ili9XXX):
+    # On the tested display the wirte direction and colormode appear to be
```
amirgon, line 593: "wirte" --> "write"

driver/esp32/ili9XXX.py
```diff
@@ -576,3 +588,104 @@ def __init__(self,
...
+        180: 0x80,
+        270: 0xE0
+    }
+
+    # GC9A01 displays handle colour mode differently than ili9XXX displays
+    COLOR_MODE = {
+        'RGB': 0x08,
+        'BGR': 0x00
+    }
+
```
amirgon, line 610: I understand that the meaning is reversed, but it's a bit confusing that for previous ili9XXX rot is set by `PORTRAIT`/`LANDSCAPE`, and for gc9a01 by number (180).  
I think it would be better to have the same notation on all controllers.  

What if gc9a01 recieved `PORTRAIT`/`LANDSCAPE`, the same way as ili9341, and translated it to the right values?
Same for `COLOR_MODE`, it's a bit confusing that colormode is a string on gc9a01 but a constant on the other controllers.

driver/esp32/ili9XXX.py
```diff
@@ -576,3 +588,104 @@ def __init__(self,
...
+        miso=5, mosi=18, clk=19, cs=13, dc=12, rst=4, power=14, backlight=15, backlight_on=0, power_on=0,
+        spihost=esp.HSPI_HOST, mhz=60, factor=4, hybrid=False, width=240, height=240,
+        colormode='RGB', rot=180, invert=False, double_buffer=True, half_duplex=True,
+        asynchronous=False, initialize=True
+    ):
+
+        if lv.color_t.SIZE != 2:
+            raise RuntimeError('gc9a01 micropython driver requires defining LV_COLOR_DEPTH=16')
+        if hybrid:
+            raise RuntimeError('gc9a01 micropython driver does not currently support hybrid driver')
```
amirgon, line 621: Did you check, maybe the C function as is (or with very small change) can work for gc9a01 as well?  
What is the difference between gc9a01 and ili9341 flush functionality? I don't see you override "flush" so I suspect [the same C "flush" function](https://github.com/lvgl/lv_binding_micropython/blob/dadad31bdce92fc96ffac6b2805ac0d5d4b884eb/driver/esp32/espidf.c#L174) ili9341 is using could be used here as well.

driver/esp32/ili9XXX.py
```diff
@@ -576,3 +588,104 @@ def __init__(self,
...
+            {'cmd': 0x67, 'data': bytes([0x00, 0x3C, 0x00, 0x00, 0x00, 0x01, 0x54, 0x10, 0x32, 0x98])},
+            {'cmd': 0x74, 'data': bytes([0x10, 0x85, 0x80, 0x00, 0x00, 0x4E, 0x00])},
+            {'cmd': 0x98, 'data': bytes([0x3e, 0x07])},
+            {'cmd': 0x35, 'data': bytes([0])},
+            {'cmd': 0x21, 'data': bytes([0])},
+            {'cmd': 0x11, 'data': bytes([0]), 'delay': 20},
+            {'cmd': 0x29, 'data': bytes([0]), 'delay': 120}
+        ]
+        
+        super().__init__(miso, mosi, clk, cs, dc, rst, power, backlight, backlight_on, power_on,
```
amirgon, line 689: Just in case we ever change any of the argument names, I recommend using named arguments for all arguments like ili9341 does:

https://github.com/lvgl/lv_binding_micropython/blob/dadad31bdce92fc96ffac6b2805ac0d5d4b884eb/driver/esp32/ili9XXX.py#L486-L491

amirgon (changes requested):

In general it looks good, I've added a few comments on your code.

driver/esp32/ili9XXX.py
minyiky: I am not sure what you mean with this one, all of the arguments in the function init are keyword parameters, the same as for the ili9341 code. If you are referring to the super().__init__ call then the ili9341 code is the same as mine

driver/esp32/ili9XXX.py
minyiky: I have no knowledge of c and will be away for a few days so didn't have time to dig around.

I tested previously and got an error but having just tried now it seems to be working, I think  had changed something somewhere else and then reverted it, I will remove this.

driver/esp32/ili9XXX.py
minyiky: I have no rechecked and it appears that portrait and landscape remain the same as the ili9341, I'm not sure why I thought it was different. I will submit a revised version.

However, is it worth allowing the option for any rotation (i.e. 0, 90, 180, 270)?

I have also added in an internal conversion for the RGB to BGR using the module defintions

driver/esp32/ili9XXX.py
minyiky: That was a mistake and has been rectified

driver/esp32/ili9XXX.py
minyiky: Thank you, now fxed

minyiky:
@amirgon thank you for the comments, I have gone through them all, made changes accordingly, and hopefully it all now makes sense.

driver/esp32/ili9XXX.py
amirgon: Are you sure you committed and pushed?  
I can still see `rot=180` on ili9488

driver/esp32/ili9XXX.py
amirgon: > However, is it worth allowing the option for any rotation (i.e. 0, 90, 180, 270)?

You can do that by adding additional constants next to `PORTRAIT` and `LANDSCAPE`.  
I'm not sure how to name them though...

driver/esp32/ili9XXX.py
amirgon: If this is working, could you change it to `hybrid=True`?
Performance improvement is significant.

driver/esp32/ili9XXX.py
embeddedt: As far as I know, reverse portrait and reverse landscape are fairly standard terms.

driver/esp32/ili9XXX.py
minyiky: Sorry I missed that, yes I will change although it will have to be Monday as I am away for the weekend

driver/esp32/ili9XXX.py
minyiky: I will check it on Monday when I get back

minyiky:
I am so sorry for the delay on this, I have now made those changes requested

amirgon:
Merged! Thank you.

We are now mainly working on [dev-8 branch](https://github.com/lvgl/lv_micropython/tree/dev-8.0) (aligned to [LVGL v8](https://blog.lvgl.io/2021-06-01/release_v8.0.0)) which will be merged to the main branch soon.  
Would you like to try this driver on dev-8 as well?

PR #145, opened 2021-05-28
/driver/generic/axp192.py: add axp192 driver

ropg:
This adds a pure python driver for the AXP192 Power Management IC


amirgon:
Merged!

Thanks.

PR #147, opened 2021-05-28
fix(driver) fix the prototype of xpt2046_read

kisvegabor:
`lv_indev_drv_t *` was missing from `STATIC bool xpt2046_read(lv_indev_data_t *data)`


amirgon:
Thanks.

@kisvegabor Why aren't you using the pure python driver [xpt2046.py](https://github.com/lvgl/lv_binding_micropython/blob/dev-8.0/driver/esp32/xpt2046.py)?
I'm not sure it makes sense to keep maintaining both C and Python drivers that do the same thing.

kisvegabor:
Actually, the issue wasn't found by me Someone reported it to me in private but he struggled with creating a PR. So I created it instead of him. 

He used the C version because the modified .c drivers gave him the ability to share the SPI bus with an SDcard.
(I don't really know more details :slightly_frowning_face:  )

PR #149, opened 2021-05-30
driver/generic/ft6x36.py: now also works with v8

ropg:
Between v7 and v8 the enum for INDEV_STATE has changed so that INDEV_STATE.PR and  INDEV_STATE.REL are now  INDEV_STATE.PRESSED and  INDEV_STATE.RELEASED respectively. I now check `lv.version_major()` and use accordingly.

(Or would you prefer two versions of the driver?)


ropg:
Thinking about it more this makes no sense: there would be comparisons all over, and clearly the right answer is to have a different driver for v8, as well as a new minimal example.

(I think this make a good argument against trivial changes like this.)

PR #165, opened 2021-07-11
Add fake lv_timer implementation for the JS port

embeddedt:
`lv_utils` is not compatible with the JS port, as it does not have a Timer implementation.

It's also not needed as I use a custom event loop which is not configurable.

### DIFF
--- a/lib/display_driver_utils.py
+++ b/lib/display_driver_utils.py
@@ -2,7 +2,8 @@ import usys as sys
 sys.path.append('') # See: https://github.com/micropython/micropython/issues/6419
 
 import lvgl as lv
-import lv_utils
+if sys.platform != 'javascript':
+    import lv_utils
 
 lv.init()
 
@@ -18,14 +19,15 @@ class driver:
         self.disp = None
         self.touch = None
         self.type = None
-        if not lv_utils.event_loop.is_running():
+        if sys.platform == 'javascript' or not lv_utils.event_loop.is_running():
             self.init_gui()
         
     def init_gui_SDL(self):
 
         import SDL
         SDL.init(w=self.width, h=self.height, auto_refresh=False)
-        self.event_loop = lv_utils.event_loop(refresh_cb = SDL.refresh)
+        if sys.platform != 'javascript':
+            self.event_loop = lv_utils.event_loop(refresh_cb = SDL.refresh)
 
         # Register SDL display driver.
 
--- a/lib/display_driver_utils.py
+++ b/lib/display_driver_utils.py
@@ -2,8 +2,7 @@ import usys as sys
 sys.path.append('') # See: https://github.com/micropython/micropython/issues/6419
 
 import lvgl as lv
-if sys.platform != 'javascript':
-    import lv_utils
+import lv_utils
 
 lv.init()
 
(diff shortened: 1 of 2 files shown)


lib/display_driver_utils.py
```diff
@@ -18,14 +19,15 @@ def __init__(self,width=420,height=320,orientation=ORIENT_PORTRAIT):
         self.disp = None
         self.touch = None
         self.type = None
-        if not lv_utils.event_loop.is_running():
+        if sys.platform == 'javascript' or not lv_utils.event_loop.is_running():
```
amirgon, line 22: Isn't that the other way around? If this is js port, we don't want to call `init_gui`.

lib/display_driver_utils.py
```diff
@@ -18,14 +19,15 @@ def __init__(self,width=420,height=320,orientation=ORIENT_PORTRAIT):
...
             self.init_gui()
         
     def init_gui_SDL(self):
 
         import SDL
         SDL.init(w=self.width, h=self.height, auto_refresh=False)
-        self.event_loop = lv_utils.event_loop(refresh_cb = SDL.refresh)
+        if sys.platform != 'javascript':
+            self.event_loop = lv_utils.event_loop(refresh_cb = SDL.refresh)
```
amirgon, line 30: If `init_gui` is not called on the js port, this is not needed.

amirgon (commented):
In general it could be cleaner to implement `Timer` for the js port, but I'm not sure how hard it is. The unix port implements it with FFI but it seems that FFI is not available for the js port.  

So if it's not easy to implement Timer, it makes sense to disable the event loop for the js port as you suggest.

lib/display_driver_utils.py
embeddedt: I think this is correct. I still use `init_gui` on the JS port to initialize SDL. The only difference is the event loop itself.

amirgon: I see, you want to prevent a call to `lv_utils.event_loop` because `lv_utils` wasn't imported...  

How about solving this differently: 

- Don't change anything in `display_driver_utils.py` or `lv_utils.py`
- Create a "dummy" Timer implementation for the JS port only.  
We could add a new JS specific `lv_timer.py` under `driver/js` and make sure `driver/js` is in the path (add it with `sys.path.append` [in the JS  port](https://github.com/lvgl/lv_micropython/blob/45c59a8f22395266c0079e115928adebe11bea2b/ports/javascript/frontend_src/main.js#L231)). It doesn't need to do anything really, just provide the Timer API.

Something like this:

```python
class Timer:

    PERIODIC = 0
    ONE_SHOT = 1

    def __init__(self, id):
        pass

    def init(self, mode=PERIODIC, period=-1, callback=None):
        pass

    def deinit(self):
        pass
```

With unchanged `display_driver_utils.py` this would create a "dummy" event loop that does nothing on the JS port, and also provide a boilerplate for a future Timer implementation for the JS port if we ever want that.  
It would also keep uniform `display_driver_utils.py` and `lv_utils.py` for all ports.

What do you think?

embeddedt: I can do this, but I'd like it to be clear to someone tinkering with the internals that the timer is a stub and does nothing, otherwise they'll wonder why the event loop doesn't work. Maybe printing a warning on the second and future times the timer is initialized is good enough. The warning would be suppressed for the [first initialization in `lv_utils`](https://github.com/lvgl/lv_binding_micropython/blob/3974a78e8effe34bbcfc22293e43398cbc2d9027/lib/lv_utils.py#L84).

amirgon: > Maybe printing a warning on the second and future times the timer is initialized is good enough

Someone tinkering with the internals might not use the simulator directly. So he would probably initialize the display+event_loop only once and won't see any warning.

We can print the warning always but it would also be displayed on the simulator, which is a bit confusing.

I think that someone tinkering with the internals would sooner or later reach driver/js/lv_timer.py, so perhaps a clear comment there could be enough?

embeddedt: I think so. I've done it this way in the latest commit. I'll merge this as it shouldn't affect anything else (just adds a new file under `driver/js`).

PR #169, opened 2021-07-23
imagetools.py: bugfix for lv.canvas()

PGNetHun:
If PNG decoder is initialized (using `imagetools.py`), then `lv.canvas()` throws `AttributeError: 'NoneType' object has no attribute '__dereference__'` exception in this line:
`png_header = bytes(lv.img_dsc_t.__cast__(src).data.__dereference__(24))`

So, basically do the following:
1. init PNG decoder
2. show a PNG image
3. initialize a canvas: `lv.canvas(lv.scr_act())`


amirgon:
Merged!

Thank you.

Issue #177, opened 2021-09-12
Consider tracking object validity


amirgon:
Consider implementing a feature that invalidates the Micropython wrapper object when LVGL object is deleted.
Related: https://forum.lvgl.io/t/how-to-delete-lvgl-objects/6770

Invalid object can have a NULL `lv_obj`.
Same C callback can be registered to all objects.
We would need to check `lv_obj != NULL` on every wrapper and throw an exception if false.

jdtsmith:
Came from the forum.  Is perhaps a simpler approach is to provide a psuedo-wrapper function like `obj.valid()` that simply checks if `lv_obj != NULL`.   I.e. provide no active validity checking on function calls, but at least expose to MP whether an object still has an LV object connected inside it.  

And it occurs to me... is there a way to accomplish this already?  I haven't found it if so.

jdtsmith:
And to expand, this simple approach would allow your MP code at least to check for and re-build LVGL objects ala:

```python
if not (self.label and self.label.valid()):
    self.label = lv.label(...)
```

Also, I presume deleting a MP wrapper object (directly or via GC) also releases the underlying LVGL object.

amirgon:
> I.e. provide no active validity checking on function calls, but at least expose to MP whether an object still has an LV object connected inside it.

In fact, the "active validity checking" is relatively cheap (check that a pointer that we reference anyway is `NULL`).  
I would rather add this validity check and throw an exception (which?), instead of expanding the API by adding another member function to each object. Some even consider [EAFP](https://docs.python.org/3/glossary.html#term-eafp) approach more "Pythonic".  

What might cost more is the event listener that should be added to each object, to detect the "delete event".

> Also, I presume deleting a MP wrapper object (directly or via GC) also releases the underlying LVGL object.

No, the MP wrapper is a reference. Deleting it (`del x`) does not delete the LVGL object, and doesn't even guarantee any resources are freed as we don't really have a "destructor" in MP that is triggered by `del`.
Calling the delete method (`x.delete()`) would delete the LVGL object.

jdtsmith:
OK, so a summary is (correct me if mistaken):

- Direction A: If the MP wrapper object is deleted, the LVGL object is not freed, and is no longer accessible (potential memory leak).
- Direction B: If the LVGL object is deleted with .delete() directly or indirectly, the MP wrapper object is not "informed" and will call into the freed memory, leading to segfaults. 

[...]

PR #178, opened 2021-09-13
Invalidate MP wrappers

amirgon:
When LVGL object is deleted, mark the Micropython object as invalid

Raise LvReferenceError when trying to access invalid object, instead of crashing

Related: https://github.com/lvgl/lv_binding_micropython/issues/177

### DIFF
--- a/gen/gen_mpy.py
+++ b/gen/gen_mpy.py
@@ -625,6 +625,7 @@ print ("""
 #include "py/binary.h"
 #include "py/objarray.h"
 #include "py/objtype.h"
+#include "py/objexcept.h"
 
 /*
  * {module_name} includes
@@ -658,6 +659,8 @@ STATIC inline const mp_obj_type_t *get_BaseObj_type()
 {{
     return &mp_lv_{base_obj}_type.mp_obj_type;
 }}
+
+MP_DEFINE_EXCEPTION(LvReferenceError, Exception)
     """.format(
             obj_type = base_obj_type,
             base_obj = base_obj_name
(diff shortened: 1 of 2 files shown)


amirgon:
@kisvegabor , @embeddedt  - what worries me about this feature is that every time a Micropython wrapper is created, it also registers an event:

https://github.com/lvgl/lv_binding_micropython/blob/ab8e1334d55c099f966e9809a7c6691258bb0ed1/gen/gen_mpy.py#L886

A Micropython wrapper object is created when a new LVGL object is created but also when converting LVGL object back to MP object. So for example, on an event handlers every call to `evt.get_target()` would register the `mp_lv_delete_cb` **again**.

Do you think it would make sense to change `lv_obj_add_event_cb` such that if the same callback function with the same user_data is registered to the same object more than once, only a single `lv_event_dsc_t` would be allocated?

kisvegabor:
> Do you think it would make sense to change lv_obj_add_event_cb such that if the same callback function with the same user_data is registered to the same object more than once, only a single lv_event_dsc_t would be allocated

To be honest I'm not completely understand the issue but wouldn't it work to add a function to LVGL the check if an event is  already added? E.g.
```c
bool lv_obj_has_event(obj, event_cb, user_data);
```

The binding could use this function to decide whether to add the the event or not. `lv_obj_has_event` can return the number of registered events with the given parameters. 


> What happens if during a tree-walk some event is fired and causes some object to be deleted?

Do you mean `lv_obj_tree_walk_cb_t` causes object deletion in some form? If so the children list of the parent or the parent itself might be corrupted. In case of a deletion `lv_obj_tree_walk_cb_t` should return `LV_OBJ_TREE_WALK_END` to stop the iteration.

amirgon:
Rethinking this - registering an event would only happen once per LVGL object, since from that moment through its lifetime it would be associated with the MP wrapper (by `user_data`). So I don't see any case where event registering would happen more than once per LVGL object, unless someone deliberately clears its `user_data`.  

Also fixed the test script to first collect objects with a tree walk and only then send the events, so no object would be deleted during the tree walk.

It seems to me stable enough.
@jdtsmith - please let me know if you find any issue.

Merging.

PR #179, opened 2021-09-17
Feat/struct subscr

amirgon:
This allows creating and accessing arrays of structs, using Python subscript operator.
Every struct actually becomes an array (with size 1 by default)

- [ ] Currently out-of-bounds access is not checked. Need to raise an exception is case of out of bounds access.

Related: https://github.com/lvgl/lvgl/pull/2565#issuecomment-922049390

### DIFF
--- a/gen/gen_mpy.py
+++ b/gen/gen_mpy.py
@@ -759,7 +759,7 @@ STATIC mp_int_t mp_lv_obj_get_buffer(mp_obj_t self_in, mp_buffer_info_t *bufinfo
 STATIC mp_int_t mp_lv_obj_get_buffer(mp_obj_t self_in, mp_buffer_info_t *bufinfo, mp_uint_t flags){ return 0; }
 #endif
 
-STATIC mp_obj_t get_native_obj(mp_obj_t *mp_obj)
+STATIC mp_obj_t get_native_obj(mp_obj_t mp_obj)
 {
     if (!MP_OBJ_IS_OBJ(mp_obj)) return mp_obj;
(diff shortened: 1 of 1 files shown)


amirgon:
Summary so far:

- Structs now have "subscript operator" that allows them to behave as C arrays.  
The "subscript operator" is equivalent to Python's `__getitem__`, `__setitem__`, `__delitem__`
- When creating a Struct, the user can specify the array size by providing an integer in the constructor to specify the number of elements. Default is 1.
- Structs can be accessed either directly (`s.field = 1`) or through the "subscript operator" (`s[17].field = 1`).  
The following are always True for Structs: `s == s[0]`, `s[0] != s[1]`, `s[10][20] == s[30]`, etc.
- `char **` are converted to `C_Pointer` Struct which provides `str_val` and therefore provides a string array (with the subscript operator)
- Arrays of C Native Types are supported only for `int` types (8bit, 16bit, 32bit and 64bit). They are represented by a new `C_Array` type which implements the "subscript operator" to access the integer values. Native arrays can be created from Python lists. Native arrays, as Structs, are convertible to Pointer (`Blob`).
- On all these new Array conversions, size is not maintained and no bound checking is performed.
Therefore `len` operator is not implemented and the array don't provide iterator (since we don't know when to stop).
- Element deletion (`del arr[17]`) simply memsets the array element at that position to 0.
- Currently `C_Array` constructor is not provided explicitly. The user would usually works with Python lists, which would be converted to `C_Array` when needed. On the other direction, a C array usually cannot be converted to Python list because its size is not known, therefore either a Struct or a `C_Array` are returned instead of a Python list when accessing a C pointer.

PR #198, opened 2021-12-03
st7789 and st7735 Support for ili9XXX hybrid driver 

russhughes:
Support ST7789 and ST7735 displays for the ESP32 LVGL MicroPython bindings. It was tested with TTGO T-Display, TTGO TWatch-2020, generic eBay ST7735, m5stack core, and core2 devices. See https://github.com/russhughes/lv_st7789 for lv_examples modified to run with this driver.


amirgon (changes requested):
Hi @russhughes ,
Thank you for making another effort and contributing these changes!   

From a quick review your changes look good. I didn't test it yet with my display but I'll get to that soon.

I think that examples are very useful for users. Do you think you could also integrate some examples from your repo? Perhaps [your "configs" scripts](https://github.com/russhughes/lv_st7789/tree/master/examples/configs) could be used as drivers for specific boards?  
A user who uses a specific board might appreciate being able to simply import a display driver script instead of configuring all the parameters to match the board.

Could you add a separate README.md file to explain the meaning and use of the png images, relation to madctl flags etc.?  
You created an `images/` directory under the lv_binding_micropython root directory. But aren't these images just examples that help with madctl configuration?  
Perhaps a better location could be under`examples/madctl` along with a README file?

driver/esp32/ili9XXX.py
```diff
@@ -56,37 +63,41 @@
...
 
-PORTRAIT = MADCTL_MX
-LANDSCAPE = MADCTL_MV
+# Negative orientation constants indicate the MADCTL value will come from the ORIENTATION_TABLE,
+# otherwise the rot value is used as the MADCTL value.
+PORTRAIT = const(-1)
+LANDSCAPE = const(-2)
+REVERSE_PORTRAIT = const(-3)
+REVERSE_LANDSCAPE = const(-4)
+ORIENTATIONS = const(4)
```
amirgon, line 81: What's the `ORIENTATIONS` constant?

driver/esp32/espidf.c
```diff
@@ -171,12 +171,16 @@ typedef struct {
 
 #define DISPLAY_TYPE_ILI9341 1
 #define DISPLAY_TYPE_ILI9488 2
+#define DISPLAY_TYPE_GC9A01  3
+#define DISPLAY_TYPE_ST7789  4
```
amirgon, line 175: Is the st7735 exactly the same as st7789?  
PR's title is "st7789 and st7735 Support" but I couldn't find references to the st7735.

driver/esp32/ili9XXX.py
russhughes: An oversight, should have been removed.

driver/esp32/espidf.h
```diff
@@ -213,16 +213,9 @@ enum {
 };
 
 enum {
```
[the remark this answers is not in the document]
russhughes: Looks like a mistake.

driver/esp32/espidf.c
russhughes: Enabling start_x and start_y parameters in the ili9341 driver allows it to work with st7735 devices.  Some st7735 devices will work with the st7789 drivers, and others won't. I suspect something in the st7789 init code disagrees with some devices. The st7735's I've tested all work with the modified ili9341 driver. They only require different parameters.

Creating an additional subclass for the st7735 with init code specific to that device might be a better solution. Either way, it should be documented better.

README.md
```diff
@@ -336,6 +336,198 @@ from ft6x36 import ft6x36
...
+  ---------- | ----- | ----------------------
+  MADCTL_MY  | 0x80  | Page Address Order
+  MADCTL_MX  | 0x40  | Column Address Order
+  MADCTL_MV  | 0x20  | Page/Column Order
+  MADCTL_ML  | 0x10  | Line Address Order
+  MADCTL_MH  | 0x04  | Display Data Latch Order
+  MADCTL_RGB | 0x00  | RGB color order
+  MADCTL_BGR | 0x08  | BGR color order
+
+### Program to help determine the `colormode`, `invert`, and `rot` init parameters
```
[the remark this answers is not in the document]
russhughes: That is a good idea.

README.md
```diff
@@ -336,6 +336,135 @@ from ft6x36 import ft6x36
...
+disp = st7789(
+    mosi=19, clk=18, cs=5, dc=27, rst=-1, backlight=12, power=-1,
+    width=240, height=240, rot=st7789.INVERSE_PORTRAIT, factor=4)
+```
+### ST7735 128x128 Configuration Example
+
+```
+from ili9XXX import ili9341, MADCTL_MX, MADCTL_MY
+
+disp = ili9341(
```
amirgon, line 435: It's a bit confusing here, creating instances of `ili9341` for ST7735 displays, users might think this is a mistake.  

As you mentioned, I also think it would be a good idea to derive new classes (`st7735_128x128`, `st7735_128x160`) from `ili9XXX` (or even from `ili9341`) with some of the arguments set to the right default values.

driver/esp32/espidf.c
```diff
@@ -171,12 +171,16 @@ typedef struct {
...
 
 void ili9xxx_flush(void *_disp_drv, const void *_area, void *_color_p)
```
amirgon, line 177: Something is wrong with the alignment in `ili9xxx_flush`.   
I think that in your changes you added tabs instead of spaces, and it looks weird when tab is configured to 4 spaces.

amirgon:
@russhughes I've reviewed your changes and tested it with a ili9341 display. Everything looks good!
What do you think about my suggestion [above](https://github.com/lvgl/lv_binding_micropython/pull/198#pullrequestreview-823298674)? :

> Perhaps [your "configs" scripts](https://github.com/russhughes/lv_st7789/tree/master/examples/configs) could be used as drivers for specific boards?
> A user who uses a specific board might appreciate being able to simply import a display driver script instead of configuring all the parameters to match the board.

driver/esp32/espidf.c
russhughes: I will look at that; it should be spaces.

README.md
russhughes: I agree it looks confusing and an st7735 class would be clearer.

russhughes:
I think that would be helpful to people, where should they go?

amirgon:
> I think that would be helpful to people, where should they go?

Maybe we can think of these "config" scripts as "Board Support Modules" - Python modules per board that import and initialize everything related to that board (usually display and input drivers, but could also be other things).
We can create a new `boards/` directory under lv_binding_micropython root and put them there. Later we could freeze them on lv_micropython for each board (with their related drivers).

One comment though. I'm not sure it's a good idea to initialize the display upon import, like you did in your "config" modules.   
It doesn't give the user a chance to configure anything. Sometimes it makes sense to configure the display differently for the same board, for example to control landscape/portrait display or use the async event loop.

README.md
amirgon: I noticed you just added `st7735` class.
Worth fixing the README.

boards/README.md
```diff
@@ -0,0 +1,43 @@
+# Board Support Modules
+
+
+## generic_st7735
+
+Module | Description
+------ | -----------
+landscape_128x128.py | st7735 128x128 display in landscape orientation |
```
amirgon, line 8: Continuing [my previous comment](https://github.com/lvgl/lv_binding_micropython/pull/198#issuecomment-988257500),  does it make sense to create different modules for landscape/portrait? Why not provide a single module the the user could import and then configure? There could be other parameters the user would want to change.

boards/lilygo_twatch_2020/twatch_landscape.py
```diff
@@ -0,0 +1,13 @@
...
+# init power manager, set backlight
+axp = axp202c.PMU()
+axp.enablePower(axp202c.AXP202_LDO2)
+axp.setLDO2Voltage(2800)
+
+# init display
+disp = st7789(
+    mosi=19, clk=18, cs=5, dc=27, rst=-1, backlight=12, power=-1,
+    width=240, height=240, start_x=0, start_y=80,
+    rot=REVERSE_LANDSCAPE, factor=4, initialize=False)
```
amirgon, line 13: What if the user wants to initialize twatch but use the async event loop, for example? Or change some other default parameter?
All is that is needed is to pass `asynchronous=True` to the display driver, but since you create the display when the module is imported the user doesn't get a chance.  

Instead, you could create a new `twatch` class with all the default parameters on the constructor and let the user instantiate it. I think that would give much more flexibility.  
The user could write something like:

```python
from twach_board import twach_driver
twatch = twatch_driver(asynchronous=True)
```

What do you think?

boards/lilygo_twatch_2020/axp202c.py
```diff
@@ -0,0 +1,820 @@
+'''
```
amirgon, line 1: Is `axp202c` a twach specific configuration?  
It looks more like a driver. Maybe it is also used by some other boards?  
A similar driver (`axp192`) is [already under `driver/generic`](https://github.com/lvgl/lv_binding_micropython/blob/master/driver/generic/axp192.py)

boards/m5stack_core2/axp202c.py
```diff
@@ -0,0 +1,820 @@
+'''
```
amirgon, line 1: Duplicated.
Also under `lilygo_twatch_2020/axp202c.py`.
Should probably go under `generic/driver`.
Is it very different from [axp192](https://github.com/lvgl/lv_binding_micropython/blob/master/driver/generic/axp192.py)?

examples/lv_example_chart_1.py
```diff
@@ -0,0 +1,54 @@
+"""
+slightly modified lv_example_chart_1.py
```
amirgon, line 2: How is this different from lv_example_chart_1.py?   
If the only difference is `import portrait_128x160 as driver` and `chart.set_size(driver.disp.width, driver.disp.height)`, then I'm not sure I see the extra benefit over lv_example_chart_1.py.

amirgon:
Merged!

If you decide to contribute the [board configuration scripts](https://github.com/lvgl/lv_binding_micropython/pull/198/commits/ebc60c2e83da32a2e5575aaf6a62aafd7bda0bd9), feel free to open another PR.

PR #205, opened 2022-02-17
Catch exceptions from task_handler in asynchronous loop

lodygaw:
Add missing try and except clauses to catch and pass exceptions from lv.task_handler() into exception_sink when using asynchronous loop.


amirgon:
Merged.

PR #210, opened 2022-04-09
generic st77xx, xpt2046 driver; rp2 DMA driver; example

eudoxos:
This PR contains

* generic pure-python driver for st77xx (7789 and 7735), with 3 models tested. The driver was written from scratch for the most part, with inspiration from drivers floating around the internet. This video only displays an arrangement of triangles, the same test is done for 3 displays, starting from the left; the other ones get artifacts, ofc, their backlight is on the same pin as I have them connected.

   rpi-displays

* generic driver for xpt2046 (touchscreen), created from scratch using datahseets and looking at various other driver;
* an example of using [Waveshare Pico-ResTouch-LCD-2.8](https://www.waveshare.com/Pico-ResTouch-LCD-2.8.htm) (st7789+xpt2046) with lvgl;
   * the display driver can optionally use rp2-specific DMA (provided, disabled by default as it might have issues and does not seem to improve the performance — did not measure, though); the DMA driver was written by @jgpeiro.
   
  Peek 2022-04-09 16-39

Thanks to @jgpeiro for the initial implementation, @amirgon for help [in this discussion](https://forum.lvgl.io/t/is-rp2-port-dysfunctional-at-runtime/6988).


driver/generic/st77xx.py
```diff
@@ -0,0 +1,404 @@
...
+            # display on
+            (ST77XX_DISPON, None, 100),
+            ## FIXME: needs out of sleep mode AGAIN, otherwise will stay bleck the first time on?
+            (ST77XX_SLPOUT, None, 100),
+        ]
+        self._run_seq(init7789)
+        self.apply_rotation(self.rot)
+
+
+if __name__=='__main__':
```
amirgon, line 339: I don't think test code belongs here.   
You can add it as a separate file under examples.  
I also think it would make more sense to provide some LVGL example instead of these rectangles.

driver/rp2/rp2_dma.py
```diff
@@ -0,0 +1,79 @@
+import time
+import uctypes
+import struct
+import machine
+
+class DMA:
+    DMA_BASE  = 0x50000000
```
amirgon, line 7: These constant should be defined as `const`?

driver/generic/xpt2046.py
```diff
@@ -0,0 +1,103 @@
...
+        spi: machine.SPI,cs,bits=12,ranges=((100,1900),(200,1950)),width=240,height=320,rot=0):
+        '''
+        Construct the Xpt2046 touchscreen controller.
+        *spi*: spi bus instance; its baud rate must *not* exceed 2_000_000 (2MHz) for correct functionality
+        *cs*: chip select (GPIO number or machine.Pin instance)
+        *bits*: ADC precision, can be 12 or 8; note that 8 will require you to provide different *ranges*
+        *ranges*: `(x_min,x_max),(y_min,y_max)` for raw coordinate readings; calibrated values might be provided.
+        *width*: width of the underyling screen in pixels, in natural (rot=0) orientation (0..*width* is the range for reported horizontal coordinate)
+        *height*: height of the underyling screen in pixels
+        *rot*: screen rotation (0: portrait, 1: landscape, 2: reverse portrait, 3: reverse landscape)
```
amirgon, line 42: Constants for `rot` orientation?

driver/generic/xpt2046.py
```diff
@@ -0,0 +1,103 @@
...
+    def _chanRead(self,chan):
+        self.cs.value(0)
+        struct.pack_into('BBB',self.buf,0,Xpt2046.START_BIT|self.conv|chan,0,0)
+        self.spi.write_readinto(self.buf,self.buf)
+        if self.conv==Xpt2046.CONV_8_BIT: ret=self.buf[1]
+        else: ret=(self.buf[1]<<4)|(self.buf[2]>>4)
+        self.cs.value(1)
+        return ret
+
+    def __init__(self,*,
```
amirgon, line 32: I think that `__init__` should also register this as an LVGL indev driver, instead of expecting the user to do it.

driver/generic/xpt2046.py
```diff
@@ -0,0 +1,103 @@
...
+        xx,yy,done=0,0,0
+        for _ in range(attempts):
+            if (r:=self._raw_pos()) is None: continue
+            xx+=r[0]; yy+=r[1]; done+=1
+            if done==N: break
+        else: return None
+        mx,my=xx*1./N,yy*1./N
+        return self._raw2px((mx,my))
+
+if __name__=='__main__':
```
amirgon, line 84: I don't think testing code should come here.  
You can add it as a separate example if you want.

driver/rp2/rp2_dma.py
```diff
@@ -0,0 +1,79 @@
...
+    HIGH_PRIO = 0x01 << 1
+    INCR_READ = 0x01 << 4
+    INCR_WRITE= 0x01 << 5
+    DREQ_PIO0_RX0 = 0x04 << 15
+    DREQ_SPI1_TX  = 0x12 << 15
+    DREQ_PERMANENT= 0x3F << 15
+    IRQ_QUIET = 0x01 << 21
+    BUSY      = 0x01 << 24
+
+    def __init__( self, channelNumber ):
```
amirgon, line 19: Should probably throw some exception if someone ever mistakenly tries to run this on platform other than rp2.

examples/generic-st7xx-with-xpt2046.py
```diff
@@ -0,0 +1,99 @@
+import sys
+sys.path.append('.')
+from st77xx import *
```
amirgon, line 3: Today most examples are assumed to be platform agnostic and are automatically run by CI.  
Your examples [fails on the CI](https://github.com/lvgl/lv_binding_micropython/runs/5955418027?check_suite_focus=true), but I think it should be excluded because it cannot run on the unix port.

To exclude an example from tests simply add it to the [EXCLUDE_PATH on `run.sh`](https://github.com/lvgl/lv_binding_micropython/blob/28109f072a05c35de67b21edf20c7300a7ee9ceb/tests/run.sh#L10).

examples/generic-st7xx-with-xpt2046.py
```diff
@@ -0,0 +1,99 @@
...
+
+print('driver')
+disp_drv=lv.disp_drv_t()
+disp_drv.init()
+
+disp_drv.draw_buf=disp_draw_buf
+disp_drv.flush_cb=disp_drv_flush_cb
+disp_drv.hor_res=lcd.width
+disp_drv.ver_res=lcd.height
+disp_drv.register()
```
amirgon, line 55: As mentioned above, I think it would be much nicer to make the display and indev driver registrations part of the driver init instead of part of the example.

examples/generic-st7xx-with-xpt2046.py
```diff
@@ -0,0 +1,99 @@
...
+lbl.set_text("Press me!")
+btn.center()
+
+btn.add_event_cb(cb_btn,lv.EVENT.CLICKED,None)
+lv.scr_load(scr)
+
+
+import lv_utils
+import uasyncio
+lv_utils.event_loop(refresh_cb=lv.task_handler,asynchronous=True)
```
amirgon, line 96: Most users don't use uasyncio.  
Maybe it's worth keeping the example simple, unless uasyncio is somehow needed for the example.

amirgon (changes requested):
Hi @eudoxos !

Thank you for this contribution! I think it's really nice to have generic platform-agnostic drivers at hand. Could be very useful not only for rp2 but also for other ports especially when someone tries some other Micropython port which we don't support with LVGL yet.

I've added some quick comments/question but didn't try to run anything yet. I don't have st7xx display at hand, but I do have xpt2046 so I'll try it later with ESP32 and see if it works for me.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,404 @@
+import time
+import machine
+import struct
+import uctypes
+
+from micropython import const
+import rp2_dma
```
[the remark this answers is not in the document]
eudoxos: This import is spurious, sorry.

driver/generic/st77xx.py
eudoxos: I can put it aside, yes.

The hardware driver is not bound to LVGL (and I think it should not be; it is indeed the problem of most python drivers for those displays that they work only for a specific framework, for not much reason TBH), and the rectangles are meant to verify that screen edges have correct offsets — this depends on the display model, and would be much more difficult to see with LVGL widgets.

This is also relevant to whether the driver should be registered with LVGL in the ctor. I'd say no, or perhaps in a derived class which would be a thin wrapper around the hardware driver itself. (And again, this wrapper could be generic and provided by LVGL, but IIRC generic driver interface was already discussed somewhere, so it is just a digression here).

driver/generic/st77xx.py
```diff
@@ -0,0 +1,404 @@
...
+    (128,160,'blacktab'):[(0,0),(0,0),(0,0),(0,0)],
+    (128,160,'redtab'):[(2,1),(1,2),(2,1),(1,2)],
+}
+
+class St77xx(object):
+    def __init__(self, *, cs, dc, bl, spi, res, suppRes, model=None, suppModel=[], rst=None, rot=1, dma=None, variant=''):
+        '''
+        * *cs*: chip select pin (= slave select, SS)
+        * *dc*: data/command pin
+        * *bl*: backlight PWM pin
```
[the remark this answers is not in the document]
eudoxos: Good point; I will make `bl` optional (i.e. `None` by default, the backlight function then being no-op).

driver/generic/st77xx.py
```diff
@@ -0,0 +1,404 @@
...
+        self.dc.value(0)
+        self.spi.write(self.buf1)
+        if buf is not None:
+            self.dc.value(1)
+            self.spi.write(buf)
+        self.cs.value(1)
+
+    def write_register_dma(self, reg, buf, is_blocking=True):
+        'If *is_blocking* is False, used should call wait_dma explicitly.'
+        SPI1_BASE = 0x40040000 # FIXME: will be different for another SPI bus?
```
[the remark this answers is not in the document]
eudoxos: It is port-specific (I might rename the function to `rp2_write_register_dma`) but it will only be called if the `dma` object (from `rp2_dma`) is provided in the ctor. I will isolate that piece somewhat more visibly, but I'd keep it in the generic file, as it is just a few lines.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,404 @@
...
+        SSPDR     = 0x008
+        self.dma.config(
+            src_addr = uctypes.addressof(buf),
+            dst_addr = SPI1_BASE + SSPDR,
```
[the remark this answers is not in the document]
eudoxos: I have no idea, this was all written by @jgpeiro and I just copied.

driver/generic/st77xx.py
amirgon: If you want a driver that is not bound to LVGL, please also provide a wrapper that is LVGL specific.   
The derived LVGL wrapper should register LVGL display driver, handle event loop etc.  This is not something we expect LVGL users to do.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,404 @@
...
+
+        if res not in suppRes: raise ValueError('Unsupported resolution %s; the driver supports: %s.'%(str(res),', '.join(str(r) for r in suppRes)))
+        if suppModel and model not in suppModel: raise ValueError('Unsupported model %s; the driver supports: %s.'%(str(model),', '.join(str(r) for r in suppModel)))
+
+        self.res=res
+        self.model=model
+
+        self.dma=dma
+        self.spi=spi
+        self.hard_reset()
```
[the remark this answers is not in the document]
eudoxos: Solved. See comment below.

eudoxos:
All the points resolved, except:

* resolution being passed as a tuple, where I just find it less confusing than width+height as that depends on orientation;
* uasync is still there, as I never used LVGL without uasync.

I added the constants for rotation even though I don't find them much useful, and they are now defined independently (thoguh identically) for `ST77XX_` and `XPT2046_`.

amirgon (changes requested):
I tested `xpt2046-test.py` with esp32 and it works without any issues.  :+1:   

I cannot test the displays, so I'll need to rely on your tests. Did you try running your display drivers on platforms other than rp2?

I've added a few more comments below.

eudoxos:
@amirgon no pressure, just let me know if there is something more to do. I will be happy if the PR is merged (even with some small TODOs for later) so that my brain can mark it as done and drop it from the hot storage :)

amirgon:
> just let me know if there is something more to do

I actually wasn't sure if you are still working on that or if you are done.  
Could you please go over our open discussions above and comment on each? (for example, you can write you implemented something as suggested, or differently, or not at all etc.)  
Then we could either continue our discussions or resolve them when we come to an agreement.

> I will be happy if the PR is merged (even with some small TODOs for later)

I would very much like to see this PR merged, but I want to merge it as cleanly as possible and not to leave TODOs for later, if we can avoid that. If we leave TODOs, there is a good chance some of them will never be addressed.  

Another challenge for me is that I'm reviewing most of this "blindly" without testing (because I'm not using rp2). The best thing could be to have one more reviewer go over your PR, someone who is familiar with rp2 and able to test your code independently.

eudoxos:
Thanks for your explanation. I added explanations to resolved conversations, still see some pending comments which I wrote though (they are marked as such by github, not sure what that means).

Outstanding points:

* should hardware-driver tests (lo LVGL involved) go to `example/test/st77xx-test.py` or to e.g. `driver/generic/st77xx-test.py`? They are an important part of the driver, for the future, as any modifications/extensions to the driver later first need to be tested on the lower level, before running LVGL on it. The internet is full of drivers which don't have such testing code and it is quite frustrating. Before I put it into `if __name__=='__main__':` of the driver and put it in separate files on your request.
* Passing the resolution as tuple or `width` / `height` separately - I left it as a tuple for now, just for my convenience. Width/height assumes orientation whether the tuple is just sorted ascending.
* There is `rp2`-specific bit in the `st77xx` generic driver (the `rp2_dma` object) which is entirely optional, even on rp2. If you are not happy with the rp2-specific bit, I am happy to drop the DMA entirely from the generic driver, as I have no knowledge of the low-level stuff to do something more with that (written by @jgpeiro; this is also related to whether DMA is limited to SPI1; I don't know)
* I added the driver registration into classes ctor, it makes actually sense and makes the app code much cleaner. I use the `lv_utils.event_loop` now in `generic-st77xx-with-2046.py` but it only works with `asynchronous=True` for reasons I don't understand (with timer-based loop, the screen shows what it should the first time but is then unresponsive).
* The driver runs with `advanced_demo` (added the platform setup in there) but does not work at all, sync or async (the screen is garbage).

eudoxos:
And yes, I understand it would be good if someone with the hardware could test and have a look, for sure. I appreciate your effort without having the hardware, really.

embeddedt:
> still see some pending comments which I wrote though (they are marked as such by github, not sure what that means).

Perhaps try to submit a review on your own PR using the "Review changes" button in the Files changed section. It seems that you have indeed left pending review comments, because we cannot see them. :slightly_smiling_face:

eudoxos (commented):
Github is smarter than I thought and I did not understand the workflow. I did not submit many (or none) of the comments here apparently. Sorry for that.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,399 @@
...
+        self._run_seq(init7789)
+        self.apply_rotation(self.rot)
+
+class St77xx_lvgl(object):
+    '''LVGL wrapper for St77xx, not to be instantiated directly.
+
+    * creates and registers LVGL display driver;
+    * allocates buffers (double-buffered by default);
+    * sets the driver callback to the disp_drv_flush_cb method.
+
```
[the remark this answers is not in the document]
amirgon: Did you remove your own event loop?  Make sure that you don't call `lv.task_handler` anywhere - it's called internally in the event loop. You don't need to set `refresh_cb`, I'll add a separate comment about this.  
Please try a minimal test script (that just displays a button [like this](https://sim.lvgl.io/v8.1/micropython/ports/javascript/index.html?script_direct=09b660dc71b3b510bd60cf5184a160b36ba8b1b1)) and see if it works. If not, please show me the script including all the imports and initialization of your drivers.

examples/generic-st77xx-with-xpt2046.py
```diff
@@ -0,0 +1,50 @@
...
+    rp2_dma=rp2_dma.DMA(0)
+else: rp2_dma=None
+
+import lvgl as lv
+lv.init()
+
+# create async event loop; this example freezes with the default sync (timer-based) event loop created in the driver (why?)
+import lv_utils
+import uasyncio
+lv_utils.event_loop(refresh_cb=lv.task_handler,asynchronous=True)
```
amirgon, line 31: You shouldn't set `refresh_cb=lv.task_handler` here.  
This is [done internally](https://github.com/lvgl/lv_binding_micropython/blob/28109f072a05c35de67b21edf20c7300a7ee9ceb/lib/lv_utils.py#L119) in event_loop.   
`refresh_cb` is usually not needed, except from cases where you need some hook to refresh the screen [for example in SDL](https://github.com/lvgl/lv_binding_micropython/blob/28109f072a05c35de67b21edf20c7300a7ee9ceb/lib/lv_utils.py#L12) but this is not needed for most display drivers.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,399 @@
...
+    def __init__(self,doublebuffer=True):
+        import lvgl as lv
+        if not lv.is_initialized(): raise RuntimeError('LVGL not initialized (call lvgl.init() first).')
+        if lv.COLOR.DEPTH!=16 or not lv.COLOR_16.SWAP: raise RuntimeError(f'LVGL *must* be compiled with 16bit color depth and swapped bytes (current: lv.COLOR.DEPTH={lv.COLOR.DEPTH}, lv.COLOR_16.SWAP={lv.COLOR_16.SWAP})')
+        disp_draw_buf=lv.disp_draw_buf_t()
+        disp_draw_buf.init(fb1:=bytearray(self.width*2*32),bytearray(self.width*2*32) if doublebuffer else None,len(fb1)//lv.color_t.__SIZE__)
+        disp_drv=lv.disp_drv_t()
+        disp_drv.init()
+        disp_drv.draw_buf=disp_draw_buf
+        disp_drv.flush_cb=lambda disp_drv,area,color: self.disp_drv_flush_cb(disp_drv,area,color)
```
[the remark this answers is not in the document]
amirgon: You can simply pass a member function. It's a bound method, so "self" will be passed along with the function object.  
Unlike C++ where you need to explicitly pass both the function pointer and "this", here it's not needed. 

[Here is an example to demonstrate this](https://sim.lvgl.io/v8.1/micropython/ports/javascript/index.html?script_direct=8eb5ecd7abdefb9e10a168f2e6f797ce3f6eee03).

examples/generic-st77xx-with-xpt2046.py
```diff
@@ -0,0 +1,46 @@
+import sys
```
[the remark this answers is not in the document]
amirgon: > Perhaps moving them to the driver itself would be better? So that there is e.g. `driver/generic/st77xx.py` and `driver/generic/st77xx-test.py`

Yes I think that would be clearer.  The "examples" directory is really for LVGL examples and not for driver test code.  
Also, you won't need to exclude each of these tests on `tests/run.sh`, which by default tests every example under "examples"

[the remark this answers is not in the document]
eudoxos: Done.

driver/generic/st77xx.py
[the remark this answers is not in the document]
eudoxos: Done, I must have been in the c++ mode when I wrote that lambda :)

driver/generic/st77xx.py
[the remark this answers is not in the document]
eudoxos: Cool, I think this did the job. Now both `advanced_demo.py` works. The minimal test script is the `examples/generic-st77xx-with-xpt2046.py`. It is significantly reduced now, no mention of event loop anywhere. So it just defines the peripherals, screen, button and that's it (it is the second video in the original PR).

examples/generic-st77xx-with-xpt2046.py
```diff
@@ -0,0 +1,46 @@
...
+btn=lv.btn(scr)
+lbl=lv.label(btn)
+lbl.set_text("Press me!")
+btn.center()
+btn.add_event_cb(cb_btn,lv.EVENT.CLICKED,None)
+lv.scr_load(scr)
+
+import lv_utils
+import uasyncio
+lv_utils.event_loop(refresh_cb=lv.task_handler,asynchronous=True)
```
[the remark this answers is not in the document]
eudoxos: I did like this in `St77xx_lvgl.__init__` (without the *asynchronous* parameter, though). It is convenient, thanks for suggesting that.
```
if not lv_utils.event_loop.is_running(): self.event_loop=lv_utils.event_loop()
```

examples/generic-st77xx-with-xpt2046.py
```diff
@@ -0,0 +1,46 @@
...
+scr=lv.obj()
+btn=lv.btn(scr)
+lbl=lv.label(btn)
+lbl.set_text("Press me!")
+btn.center()
+btn.add_event_cb(cb_btn,lv.EVENT.CLICKED,None)
+lv.scr_load(scr)
+
+import lv_utils
+import uasyncio
```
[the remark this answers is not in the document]
eudoxos: Done, this is much simplified now as you suggested, without asyncio.

examples/generic-st77xx-with-xpt2046.py
eudoxos: I removed all that setup from the `generic-st77xx-with-xpt2046.py` and it works better now.

eudoxos:
`advanced_demo.py` works nicely now. Just the colors seem to be wrong (blue shows red gradient). Is it because I compiled with `LV_COLOR_16_SWAP` and should not, for this hardware? 

image

examples/generic-st77xx-with-xpt2046.py
eudoxos: Done as you suggested. It works better now :)

amirgon:
> Just the colors seem to be wrong (blue shows red gradient, green shows green, red shows blue). Is it because I compiled with `LV_COLOR_16_SWAP` and should not, for this hardware?

The color distortion issue needs to be addressed.  
`LV_COLOR_16_SWAP` is [explained in the docs](https://docs.lvgl.io/master/overview/color.html?highlight=lv_color_16_swap), and is usually needed when sending 16-bit values through SPI interface.  

Your driver should verify the correct settings of `LV_COLOR_DEPTH` and `LV_COLOR_16_SWAP` and throw an error if they are incorrectly set. See an example of doing that in [ili9XXX driver](https://github.com/lvgl/lv_binding_micropython/blob/28109f072a05c35de67b21edf20c7300a7ee9ceb/driver/esp32/ili9XXX.py#L546-L549).

Color distortion may be a result of an error in your driver init sequence. For example selecting RGB/ BGR order. Please consult with the display datasheet.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,405 @@
...
+
+        if lv.COLOR.DEPTH!=16 or not lv.COLOR_16.SWAP: raise RuntimeError(f'LVGL *must* be compiled with 16bit color depth and swapped bytes (current: lv.COLOR.DEPTH={lv.COLOR.DEPTH}, lv.COLOR_16.SWAP={lv.COLOR_16.SWAP})')
+
+        if not lv.is_initialized(): lv.init()
+        # create event loop if not yet present
+        if not lv_utils.event_loop.is_running(): self.event_loop=lv_utils.event_loop()
+
+        # attach all to self to avoid objects' refcount dropping to zero when the scope is exited
+        self.disp_draw_buf=lv.disp_draw_buf_t()
+        self.disp_draw_buf.init(fb1:=bytearray(self.width*2*32),bytearray(self.width*2*32) if doublebuffer else None,len(fb1)//lv.color_t.__SIZE__)
```
amirgon, line 385: Could you explain the `*2*32` multiplication?  How did you decide on this buffer size?  
Why not let the user control the buffer size? On ili9XXX we use `factor` to allow the user control how much of the display buffer should be allocated. (factor 4 means allocate only 1/4 of the display size, and update the display in 8 parts).

driver/generic/xpt2046.py
```diff
@@ -0,0 +1,123 @@
...
+        else: (data.point.x,data.point.y),data.state=pos,1
+        # print('#',end='')
+        # switch SPI back to 24 MHz for the display
+        if self.spiRate: self.spi.init(baudrate=self.spiRate)
+        return False
+
+    def __init__(self,spi,spiRate=24_000_000,lcd=None,**kw):
+        '''XPT2046 touchscreen driver for LVGL; cf. documentation of :obj:`Xpt2046_hw` for the meaning of parameters being passed.
+
+        *lcd*: optional St77xx instance; in case of port-specific async DMA transfer on RP2, it will wait for the transfer to finish before using the SPI bus.
```
amirgon, line 109: xpt2046 is not really related to st77xx and can used with other displays, so I don't think it's clean to pass st77xx lcd parameter here. An input device driver should be completely independent of the display driver.  

I understand you want to wait until DMA transfer is finished, and this doesn't have to be specific to st77xx. Instead of passing the lcd you could pass an optional callback parameter that waits for SPI to be ready after DMA finished (something like `wait_for_spi` that returns when SPI is ready). Such callback could potentially be useful with other displays in the future.

driver/generic/xpt2046.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class Xpt2046(Xpt2046_hw):
+    lcd=None
+    spi=None
+    def indev_drv_read_cb(self, indev_drv, data):
+        # wait for DMA transfer (if any) before switchint SPI to 1 MHz
+        if self.lcd: self.lcd._rp2_wait_dma()
+        # print('.',end='')
+        if self.spiRate: self.spi.init(baudrate=1_000_000)
```
amirgon, line 97: Do you initialize spi on every call to `indev_drv_read_cb`?  
Why not do that on the constructor?

driver/generic/xpt2046.py
```diff
@@ -0,0 +1,123 @@
...
+        pos=self.pos()
+        if pos is None: data.state=0
+        else: (data.point.x,data.point.y),data.state=pos,1
+        # print('#',end='')
+        # switch SPI back to 24 MHz for the display
+        if self.spiRate: self.spi.init(baudrate=self.spiRate)
+        return False
+
+    def __init__(self,spi,spiRate=24_000_000,lcd=None,**kw):
```
amirgon, line 106: Why do you set default spiRate to 24_000_000 here?

driver/generic/xpt2046.py
```diff
@@ -0,0 +1,123 @@
...
+    def indev_drv_read_cb(self, indev_drv, data):
+        # wait for DMA transfer (if any) before switchint SPI to 1 MHz
+        if self.lcd: self.lcd._rp2_wait_dma()
+        # print('.',end='')
+        if self.spiRate: self.spi.init(baudrate=1_000_000)
+        pos=self.pos()
+        if pos is None: data.state=0
+        else: (data.point.x,data.point.y),data.state=pos,1
+        # print('#',end='')
+        # switch SPI back to 24 MHz for the display
```
amirgon, line 102: I don't think you should assume the display is necessarily 24 MHz.

driver/generic/xpt2046.py
jgpeiro: It might seem a bit weird at first, but on the board I used, the TSC and the LCD were connected to the same SPI. The LCD works at full speed (>10MHz), but the TSC doesn't work faster than (about 1MHz), so I had to reconfigure it every time I wanted to read the TSC. It is also necessary to wait for the DMA to finish to reconfigure the SPI for same reason.

If I had designed the hardware, I would have preferred a dedicated SPI for the LCD and a more "independent" sw, but many boards share it since the RP2040 doesn't have too many pins.

driver/generic/xpt2046.py
jgpeiro: It's not a good practice, but it was the only solution because, as I wrote before, the hardware used shares the SPI bus, so indev and the display driver are "dependent" on the HW side.

I also don't like calling spi.init multiple times, but based on my experience with micropython, spi.init should reset all peripheral registers and work as expected. I only tested with RP2, and I think it should also work on STM32 and ESP32, but I'm not sure and I know that it might be something that could cause problems for other platforms.

driver/generic/xpt2046.py
eudoxos: As a side note, there does not seem to be API to query the *current* SPI rate (to read it, set to 1MHz for reading from xpt2046, then set it back), that's why `Xpt2046.spiRate` is needed to store the desired non-touch SPI rate.

driver/generic/st77xx.py
eudoxos: Added *factor*, thanks for the suggestion (`2*32` rows was 32 rows with 2 bytes per pixel).

driver/generic/st77xx.py
```diff
@@ -0,0 +1,405 @@
...
+        # print(f"({area.x1},{area.y1}..{area.x2},{area.y2})")
+        self._rp2_wait_dma() # wait if not yet done and DMA is being used
+        # blit in background
+        self.blit(area.x1,area.y1,w:=(area.x2-area.x1+1),h:=(area.y2-area.y1+1),disp_drv.draw_buf.buf_act.__dereference__(2*w*h),is_blocking=False)
+        self.disp_drv.flush_ready()
+    def __init__(self,doublebuffer=True):
+        import lvgl as lv
+        import lv_utils
+
+        if lv.COLOR.DEPTH!=16 or not lv.COLOR_16.SWAP: raise RuntimeError(f'LVGL *must* be compiled with 16bit color depth and swapped bytes (current: lv.COLOR.DEPTH={lv.COLOR.DEPTH}, lv.COLOR_16.SWAP={lv.COLOR_16.SWAP})')
```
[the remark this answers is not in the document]
eudoxos: Done, the class now support `bgr=True|False` and checks color depth and swap accordingly. As a side note, I would suggest to expose that in a friendlier manner in python (not immediately clear what `hasattr(lv.color_t().ch,'green_l')` means).

driver/generic/xpt2046.py
eudoxos: I am passing `spiPrereadCb` now, good idea.

driver/generic/xpt2046.py
eudoxos: There were comments by @jgpeiro here which I don't see now (??). The St77xx display I happen to have shares the SPI bus with the Xpt2046 driver and they both need different frequencies. The St77xx woule run at 1MHz but that would be terribly slow, the Xpt2046 cannot handle more than about 1MHz. So one needs to switch SPI to lower speed before reading. I did not find another way to do it than initing the bus with a different baudrate, that seems to work nicely.

`Xpt2046.spiRate` is the rate the SPI bus should be *restored to* after the touch driver is done reading. If `spiRate` is not given at all, then there is no rate-switching on the SPI bus.

driver/generic/xpt2046.py
eudoxos: As explained above, that is the rate to return the bus to (Xpt2046 always sets it to 1MHz because of chip limitations).

driver/generic/xpt2046.py
eudoxos: This was an old comment which I removed; the new rate will be Xpt2046.spiRate.

eudoxos:
> The color distortion issue needs to be addressed. LV_COLOR_16_SWAP is [explained in the docs](https://docs.lvgl.io/master/overview/color.html?highlight=lv_color_16_swap), and is usually needed when sending 16-bit values through SPI interface.

This is fixed now.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,415 @@
...
+        self.buf1 = bytearray(1)
+        self.buf2 = bytearray(2)
+        self.buf4 = bytearray(4)
+
+        self.cs,self.dc,self.bl,self.rst=[(machine.Pin(p,machine.Pin.OUT) if isinstance(p,int) else p) for p in (cs,dc,bl,rst)]
+        self.bl.value(1)
+        self.bl=machine.PWM(self.bl)
+        self.rot=rot
+        self.bgr=bgr
+        self.width,self.height=(0,0) # set in apply_rotation
```
amirgon, line 146: What is the motivation in calling `apply_rotation` on each subclass instead of once on the base class? (probably right after calling `config`)

driver/generic/st77xx.py
```diff
@@ -0,0 +1,415 @@
...
+
+        self.res=res
+        self.model=model
+
+        self.rp2_dma=rp2_dma
+        self.spi=spi
+        self.hard_reset()
+
+    def off(self):
+        self.bl.value(0)
```
amirgon, line 159: Even if `bl` is not `None` - this would still crash because you've set `self.bl` to `machine.PWM` which doesn't have "value" function.  
If you want PWM at least name it differently instead of re-assigning `self.bl` to a different type. This would help prevent confusion.

examples/generic-st77xx-with-xpt2046.py
```diff
@@ -0,0 +1,36 @@
...
+    # with DMA, the repaints seem to be too slow? To be investigated
+    # we seem to be fine performance-wise without DMA with 320x240 anyway
+    import rp2_dma
+    rp2_dma=rp2_dma.DMA(0)
+else: rp2_dma=None
+
+import lvgl as lv
+lv.init()
+
+lcd=St7789(rot=3,res=(240,320),spi=spi,cs=9,dc=8,bl=13,rst=15,rp2_dma=rp2_dma,factor=8)
```
amirgon, line 26: You've already added St7789 and Xpt2046 to advanced_demo example - which is good.  
Is there a need for this tiny example? What does it add over the example code in advanced_demo?  
If you agree this does not add much, consider removing it and the revert run.sh.

driver/generic/st77xx.py
eudoxos: The motivation was that apply_rotation manipulates the registers, thus the logical place for it was at the end of each subclass's `config` which does the first part of the hardware setup. I could also create a method `St77xx_hw.config` calling the real init sequence as e.g. `St7735_hw.config_hw`  and then `St77xx.apply_rotation`.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,415 @@
...
+
+        * *suppModel*: models supported by the hardware driver
+        * *suppRes*: resolutions supported by the hardware driver, as list of (width,height) tuples
+        '''
+        self.buf1 = bytearray(1)
+        self.buf2 = bytearray(2)
+        self.buf4 = bytearray(4)
+
+        self.cs,self.dc,self.bl,self.rst=[(machine.Pin(p,machine.Pin.OUT) if isinstance(p,int) else p) for p in (cs,dc,bl,rst)]
+        self.bl.value(1)
```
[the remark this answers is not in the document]
eudoxos: Thanks for catching this.

driver/generic/st77xx.py
eudoxos: Same as above, thank's for catching that. I will test with `bl=None` later.

examples/generic-st77xx-with-xpt2046.py
eudoxos: Not that I can find the comment right now, but you requested a minimal example with a button, which is this one. At first, I had no idea how to run `advanced_demo` with rp2 so it was the LVGL test for me. It makes the relatively simple setup stand out (though I understand it might be superfluous) plus there is the `rp2_dma` portion which someone might tune later (it works, but less efficiently than without).

driver/generic/st77xx.py
amirgon: If you always set these registers the same way on all hw flavours, I don't see the benefit in duplicating code.  

In fact, you could simply pass the init array to the St77xx_hw constructor (in `super().__init__` ) and let St77xx_hw call both `_run_seq` and `apply_rotation`, moving all functionality to St77xx_hw and keeping only configuration data on the derived classes.

driver/generic/st77xx.py
```diff
@@ -0,0 +1,417 @@
...
+
+        * *suppModel*: models supported by the hardware driver
+        * *suppRes*: resolutions supported by the hardware driver, as list of (width,height) tuples
+        '''
+        self.buf1 = bytearray(1)
+        self.buf2 = bytearray(2)
+        self.buf4 = bytearray(4)
+
+        self.cs,self.dc,self.rst=[(machine.Pin(p,machine.Pin.OUT) if isinstance(p,int) else p) for p in (cs,dc,rst)]
+        if isinstance(self.bl,int): self.bl=machine.PWM(self.bl)
```
amirgon, line 142: I believe that in most Micropython ports PWM expects a `machine.Pin` object and not an `int`.  
See [ESP32 PWM for example](https://docs.micropython.org/en/latest/esp32/tutorial/pwm.html#pulse-width-modulation).

I'm not sure why it works with an `int` on rp2 (did you test this?)

driver/generic/st77xx.py
```diff
@@ -0,0 +1,417 @@
+import time
+import machine
+import struct
+import uctypes
+
+from micropython import const
+
+#
+# This driver was written from scratch using datasheets and looking at other drivers listed here:
```
amirgon, line 9: When you base your code on code of others, please pay attention to the legal aspects:
- Make sure you copy only from sources with permissive license (such as MIT, BSD etc.) and not GPL or LGPL or others which are incompatible with LVGL permissive (MIT) license. 
- Make sure you include the original license when needed.

For example, Adafruit ST7789 license is MIT, but [requires to include the copyright notice](https://github.com/pimoroni/st7789-python/blob/3e63dc34042c0d12bf2950ed5b88d3a8dc060018/library/ST7789/__init__.py#L11-L12):
```
# The above copyright notice and this permission notice shall be included in
# all copies or substantial portions of the Software.
```

driver/generic/st77xx.py
eudoxos: I created `config_hw` in child classes based on my previous suggestion. Putting everything to St77xx_hw has its pros, but also cons: it will duplicate supported resolution / model handling in the child class ctors (perhaps others).

driver/generic/st77xx.py
eudoxos: `machine.PWM(int)` [is possibly prot-specific](https://docs.micropython.org/en/latest/library/machine.PWM.html#machine.PWM) so I handle that manually now.

driver/generic/st77xx.py
eudoxos: Done in the last commit. The Xpt2046 is not "based on" the driver linked, it was just an inspiration (like most of the others, anyway).

amirgon:
Merged!

Thank you very much for this contribution.

PR #211, opened 2022-05-12
added SDL keyboard handler

daviel:
I was needing a keyboard driver using SDL under an unix system and thought that it could be useful for others in the future.

### DIFF
--- a/driver/SDL/SDL_keyboard.c
+++ b/driver/SDL/SDL_keyboard.c
@@ -1,0 +1,150 @@
+/**
+ * @file keyboard.c
+ *
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "SDL_keyboard.h"
+#if USE_KEYBOARD != 0
+
+/*********************
+ *      DEFINES
+ *********************/
+#ifndef MONITOR_ZOOM
+#define MONITOR_ZOOM    1
+#endif
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+static bool keyup = false;
+static bool keydown = false;
+static bool repeat = false;
+static bool keyboard_was_read = true;
+static int last_key = -1;
+
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+/**
+ * Get the current key and state of the keyboard
+ * @param data store the keyboard data here
+ * @return false
+ */
+void keyboard_read(struct _lv_indev_drv_t * indev_drv, lv_indev_data_t * data)
+{
+    if(!keyboard_was_read)
+    {
+        /* Store the collected data */
+        if(keyup)
+        {
+            data->state = LV_INDEV_STATE_PR;
+            data->key = last_key;
+        }
+        else if(keydown)
+        {
+            data->state = LV_INDEV_STATE_REL;
+            data->key = last_key;
+        }
+    }
+    if(keydown && repeat){
+        data->state = LV_INDEV_STATE_PR;
+        data->key = last_key;
+    }
+    
+    data->continue_reading = false;
+    keyboard_was_read = true;
+    return;
+}
+
+/**
+ * It will be called from the main SDL thread
+ */
+void keyboard_handler(SDL_Event * event)
+{
+    keydown = false;
+    keyup = false;
+    last_key = -1;
+
+    if(keyboard_was_read)
+    {
+        switch(event->type) {
+            case SDL_KEYUP:
+                    keyup = true;
+                    repeat = event->key.repeat;
+                    keyboard_was_read = false;
+                break;
+            case SDL_KEYDOWN:
+                    keydown = true;
+                    repeat = event->key.repeat;
+                    keyboard_was_read = false;
+                break;
+        }
+
+        if(keyup || keydown)
+        {
+            switch(event->key.keysym.sym)
+            {
+                case SDLK_TAB:
+                    last_key = LV_KEY_NEXT;
+                    break;
+                case SDLK_END:
+                    last_key = LV_KEY_PREV;
+                    break;
+                case SDLK_RETURN:
+                    last_key = LV_KEY_ENTER;
+                    break;
+                case SDLK_LEFT:
+                    last_key = LV_KEY_LEFT;
+                    break;
+                case SDLK_RIGHT:
+                    last_key = LV_KEY_RIGHT;
+                    break;
+                case SDLK_UP:
+                    last_key = LV_KEY_UP;
+                    break;
+                case SDLK_DOWN:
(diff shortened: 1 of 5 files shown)


driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * It will be called from the main SDL thread
+ */
+void keyboard_handler(SDL_Event * event)
+{
+    keydown = false;
+    keyup = false;
+    last_key = -1;
+
+    if(keyboard_was_read)
```
amirgon, line 85: Keys are recorded only when `keyboard_was_read` is true. `keyboard_was_read` is set to true only after `keyboard_read` was called by LVGL, and set to false by `keyboard_handler`.  

Suppose `keyboard_handler` is called more frequently than `keyboard_read`. This means that keys are recorded on the **first** invocation of `keyboard_handler` after a call to `keyboard_read` and not updated until the next call to `keyboard_read`. If keyboard state changes, that would not be recorded (until the next call to `keyboard_read` followed by a call to `keyboard_handler`).  

Is it designed that way on purpose?   
Perhaps it would make more sense to record the keyboard state on the **last** invocation of `keyboard_handler` before `keyboard_read` is called?

driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+ **********************/
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+static bool keyup = false;
+static bool keydown = false;
+static bool repeat = false;
+static bool keyboard_was_read = true;
+static int last_key = -1;
```
amirgon, line 34: These key state globals are accessed from different threads right? The SDL thread and Micropython thread.  
How do you ensure atomic updates of this state? (that all of the key state globals are written before being read by the other thread)

driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+/**
+ * Get the current key and state of the keyboard
+ * @param data store the keyboard data here
+ * @return false
```
amirgon, line 48: Does not return false, in fact.

driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+    if(keydown && repeat){
+        data->state = LV_INDEV_STATE_PR;
+        data->key = last_key;
+    }
+    
+    data->continue_reading = false;
+    keyboard_was_read = true;
+    return;
```
amirgon, line 73: meaningless `return` at the end of void function.

driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+        }
+        else if(keydown)
+        {
+            data->state = LV_INDEV_STATE_REL;
+            data->key = last_key;
+        }
+    }
+    if(keydown && repeat){
+        data->state = LV_INDEV_STATE_PR;
+        data->key = last_key;
```
amirgon, line 68: If you update `data->key = last_key` the same way in all cases, you can move it outside the "if" block.

driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+        switch(event->type) {
+            case SDL_KEYUP:
+                    keyup = true;
+                    repeat = event->key.repeat;
+                    keyboard_was_read = false;
+                break;
+            case SDL_KEYDOWN:
+                    keydown = true;
+                    repeat = event->key.repeat;
+                    keyboard_was_read = false;
```
amirgon, line 96: `repeat` and `keyboard_was_read` are updated the same way on both cases - you can move them out of the "switch" block.

driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+                break;
+        }
+
+        if(keyup || keydown)
+        {
+            switch(event->key.keysym.sym)
```
amirgon, line 102: I think it would be cleaner to have a seperate function do the translation from `event->key.keysym.sym` to LVGL key.  
See [keycode_to_ctrl_key](https://github.com/lvgl/lv_drivers/blob/75088aecdb7a90d27152b49e9917c90611471c7f/sdl/sdl_common.c#L223) as an example.

driver/SDL/SDL_keyboard.h
```diff
@@ -0,0 +1,73 @@
...
+#ifdef LV_CONF_INCLUDE_SIMPLE
+#include "lv_drv_conf.h"
+#else
+#include "lv_drv_conf.h"
+#endif
+
+#if USE_KEYBOARD
+#include <stdint.h>
+#include <stdbool.h>
+#include <SDL2/SDL.h>
```
amirgon, line 25: `SDL2/SDL.h` is already included below, with `#include MONITOR_SDL_INCLUDE_PATH`.

driver/SDL/SDL_keyboard.h
```diff
@@ -0,0 +1,73 @@
...
+ **********************/
+
+/**
+ * Initialize the keyboard
+ */
+void keyboard_init(void);
+/**
+ * Get the current position and state of the keyboard
+ * @param data store the keyboard data here
+ * @return false: because the points are not buffered, so no more data to be read
```
amirgon, line 54: Does not return false.

driver/SDL/SDL_keyboard.c
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+    return;
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+#endif
```
amirgon, line 150: `#endif /* USE_KEYBOARD */`

amirgon (changes requested):
Hi @daviel, thank you for your contribution!

Are you aware of the [LVGL SDL driver under lv_drivers](https://github.com/lvgl/lv_drivers/tree/master/sdl)? Perhaps it's best to base on that since it's well maintained and tested . For example, I noticed it supports some keys which are missing on your PR (see [keycode_to_ctrl_key](https://github.com/lvgl/lv_drivers/blob/75088aecdb7a90d27152b49e9917c90611471c7f/sdl/sdl_common.c#L223))

The SDL Micropython driver was written long time ago and not well maintained (but it works). The code under lv_drivers is newer, well maintained and much more tested. It would probably be best to simply replace the entire SDL Micropython driver by one which is based on lv_drivers, instead of patching the exiting Micropython driver.

Anyway, I'm adding a few questions/comments on your code.

amirgon:
Superseded by https://github.com/lvgl/lv_binding_micropython/pull/215

daviel:
> Superseded by #215

Thank you very much for your work. I did not have the time to rewrite MR so I am very happy to test your solution. :)

amirgon:
> so I am very happy to test your solution. :)

Great! Let me know if you see any problem.

PR #225, opened 2022-07-25
Align to LVGL v9 updates

amirgon:
Related: https://github.com/lvgl/lvgl/pull/3501

### DIFF
--- a/driver/esp32/ili9XXX.py
+++ b/driver/esp32/ili9XXX.py
@@ -97,7 +97,8 @@ class ili9XXX:
     def __init__(self,
         miso=5, mosi=18, clk=19, cs=13, dc=12, rst=4, power=14, backlight=15, backlight_on=0, power_on=0,
         spihost=esp.HSPI_HOST, spimode=0, mhz=40, factor=4, hybrid=True, width=240, height=320, start_x=0, start_y=0,
-        invert=False, double_buffer=True, half_duplex=True, display_type=0, asynchronous=False, initialize=True
+        invert=False, double_buffer=True, half_duplex=True, display_type=0, asynchronous=False, initialize=True,
+        color_format=lv.COLOR_FORMAT.NATIVE
     ):
 
         # Initializations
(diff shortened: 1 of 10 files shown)


driver/esp32/ili9XXX.py
```diff
@@ -97,7 +97,8 @@ class ili9XXX:
     def __init__(self,
         miso=5, mosi=18, clk=19, cs=13, dc=12, rst=4, power=14, backlight=15, backlight_on=0, power_on=0,
         spihost=esp.HSPI_HOST, spimode=0, mhz=40, factor=4, hybrid=True, width=240, height=320, start_x=0, start_y=0,
-        invert=False, double_buffer=True, half_duplex=True, display_type=0, asynchronous=False, initialize=True
+        invert=False, double_buffer=True, half_duplex=True, display_type=0, asynchronous=False, initialize=True,
+        color_format=lv.COLOR_FORMAT.NATIVE
```
kisvegabor, line 101: `NATIVE` is the default, it's not required to set it explicitly.

kisvegabor (commented):
Looks good, only one comment. 

Could you test it with ESP? If so, was there any significant change in performance due to `LV_COLOR_FORMAT_NATIVE_REVESE`?

examples/custom_widget_example.py
```diff
@@ -130,7 +130,9 @@ def draw(self, obj, draw_ctx):
 class CustomWidget():
 
     # An instance of a widget-class to be used for creating custom widgets
-    cls = CustomWidgetClass(lv.DPI.DEF, lv.DPI.DEF)
+    d = lv.disp_get_default();
```
embeddedt, line 133: Minor nitpick, but the semicolon is unnecessary here, right?

embeddedt (approved):

amirgon:
> Could you test it with ESP? If so, was there any significant change in performance due to `LV_COLOR_FORMAT_NATIVE_REVESE`?

I didn't notice significant performance issues on ESP32.

driver/esp32/ili9XXX.py
amirgon: Replaced with default `None`

examples/custom_widget_example.py
amirgon: Right.  
Actually there were lots of spurious semicolons...  
I now removed those I found.

kisvegabor:
> I didn't notice significant performance issues on ESP32.

Great!

PR #255, opened 2023-02-06
Update pycparser to v2.21

Sh3Rm4n:
See https://github.com/lvgl/lvgl/pull/3965


gen/gen_mpy.py
```diff
@@ -674,6 +675,7 @@ def register_int_ptr_type(convertor, *types):
 
 #include <stdlib.h>
 #include <string.h>
+#include <stdatomic.h>
```
amirgon, line 678: Is `<stdatomic.h>` always supported?
What if the user is using an older compiler, or an architecture where atomics are not available?
What if the compiler is C99 and not C11?

amirgon (changes requested):
Hi @Sh3Rm4n, thank you for this contribution.
If you taking the opportunity to update pycparser, why not update it to latest version?

gen/gen_mpy.py
Sh3Rm4n: Yeah, I'm unsure about that as well. This is at least needed for now, to be able to compile `lv_mpy.c`, though I share your concerns. I wonder why it even compiles. As far as I can see, the makefile set's `-std=gnu99`. Maybe `gnu99` supports `stdatomic` already? 

Anyway, the include is needed, because with the pycparser update, the [`memory_order`](https://github.com/eliben/pycparser/blob/3cf6bf5eb16f5eadd4a058e41596145c407a79ad/utils/fake_libc_include/_fake_typedefs.h#L213) enum appears in the preprocessed file `gen_mpy.py` takes. And `gen_mpy` converts that enum. If `stdatomic.h` is not included, the compiler can not find the enum variants. 

I think, that `gen_mpy.py` does not need to convert the `memory_order` enum flags, but I do not now, where to exclude it. But as I don't know much about the `gen_mpy` process, you might now a better solution :) 

Regarding the pycparser version, I've updated to the `v2.21` release, assuming that this is the latest stable release, which also adds the atomic support I need, but if it is okay for you, I'll update to the latest master.

amirgon: > Anyway, the include is needed, because with the pycparser update, the [`memory_order`](https://github.com/eliben/pycparser/blob/3cf6bf5eb16f5eadd4a058e41596145c407a79ad/utils/fake_libc_include/_fake_typedefs.h#L213) enum appears in the preprocessed file `gen_mpy.py` takes. And `gen_mpy` converts that enum. If `stdatomic.h` is not included, the compiler can not find the enum variants.

`gen_mpy.py` generates `lv_mpy.c` with the binding code.  It can be used with LVGL to create the LVGL bindings, but can also be used with other libraries unrelated to LVGL. But `stdatomic.h` is only needed for LVGL so I don't think it belongs in  `gen_mpy.py`.  
When generating bindings for LVGL, the generated `lv_mpy.c` already includes `lvgl.h`.  
So if `stdatomic.h` was included in `lvgl.h`, it would be processed. I don't understand why you want to include `stdatomic.h` explicitly in `gen_mpy.py`.

Are you planing to always require atomic support when building LVGL? Why not support it only optionally with some macro to enable/disable it on lv_conf.h?

> Regarding the pycparser version, I've updated to the `v2.21` release, assuming that this is the latest stable release, which also adds the atomic support I need, but if it is okay for you, I'll update to the latest master.

It's ok to support v2.21 and not latest.

Sh3Rm4n: > But stdatomic.h is only needed for LVGL so I don't think it belongs in gen_mpy.py.

Yes, that's what I'm thinking as well. I just had to include `stdatomic.h` so that `lv_mpy.c` builds again, the pycparser update introduces the `memory_order` enums unconditionally.

> When generating bindings for LVGL, the generated lv_mpy.c already includes lvgl.h.
So if stdatomic.h was included in lvgl.h, it would be processed. I don't understand why you want to include stdatomic.h explicitly in gen_mpy.py.

Interesting. It seems like these change blocks . Currently, `stdatomic.h` is not included at all, so `lv_mpy.c` does not build, if https://github.com/lvgl/lvgl/pull/3965 is merged it will be (as long as the compiler supports it and `C11` is used as the c-std). But I've updated pycparser to micropython build again for that PR. So until that PR is merged, `stdatomic.h` is not included at all. 

> Are you planing to always require atomic support when building LVGL? Why not support it only optionally with some macro to enable/disable it on lv_conf.h?

This is an ongoing discussion and also my first approach in https://github.com/lvgl/lvgl/pull/3965, but we are currently settling on enabling atomic support by default, if the compiler supports it: https://github.com/lvgl/lvgl/pull/3965#issuecomment-1427539148

So I think it is not fully solved by an `stdatomic.h` include of `lvgl.h`,  as this currenlty not done and probably done conditionally by the compiler. The `memory_model` enum introduced by the pycparser update is however always added unconditionally. 
But I agree, that we do not need bindings for this enum. Not now, nor in the future, as bindings for atomic variables are not needed (and also do not make sense AFAIK). 

What I would propose is to add some code go `gen_mpy.py` to specifically exclude this enum? Solving this issue: 
https://github.com/lvgl/lv_micropython/pull/62/checks?check_run_id=11131973194

```
/home/runner/work/lv_micropython/lv_micropython/ports/esp32/build-GENERIC/lv_mp.c:3412:59: error: 'memory_order_relaxed' undeclared here (not in a function); did you mean 'lv_obj_del_delayed'?
     { MP_ROM_QSTR(MP_QSTR_relaxed), MP_ROM_PTR(MP_ROM_INT(memory_order_relaxed)) },
```

or provide another way so that `stdatomic.h` is not spilled into the `lv_mpy.c` for generating bindings: https://github.com/lvgl/lvgl/actions/runs/4101455929/jobs/7073257529

> It's ok to support v2.21 and not latest.

:+1:

Sh3Rm4n: Alright, so here is the compromise I was talking about: I do not include `stdatomic.h` but also introduce a special case for the `memory_order` enum, so no bindings are generated, see [`8572bb4` (#255)](https://github.com/lvgl/lv_binding_micropython/pull/255/commits/8572bb4671a61a7c74de1ec45616ec8a1de949c0)

amirgon: > but also introduce a special case for the `memory_order` enum, so no bindings are generated, see [`8572bb4` (#255)](https://github.com/lvgl/lv_binding_micropython/pull/255/commits/8572bb4671a61a7c74de1ec45616ec8a1de949c0)

I'm currently traveling so I can't test this, but were is `memory_order` enum defined?  Why isn't the header it's defined in included together with stdatomic.h? I don't worry too much about an extra enum,  but if anyone ever wanted this enum generated, either with LVGL or when generating bindings for some other library, you just blocked that option...

Sh3Rm4n: The [`memory_order`](https://github.com/eliben/pycparser/blob/3cf6bf5eb16f5eadd4a058e41596145c407a79ad/utils/fake_libc_include/_fake_typedefs.h#L213) enum is defined and introduced by the pycparser [`_fake_typedefs.h`](https://github.com/eliben/pycparser/blob/3cf6bf5eb16f5eadd4a058e41596145c407a79ad/utils/fake_libc_include/_fake_typedefs.h) file, but is normally defined in the `stdatomic.h` 

Because of the way pycparser collects all these typedefs in one file, this enum is implicitly included everywhere, even though `stdatomic.h` is not included. 

This results into it appearing in the symbol file `gen_mpy.py` uses to generate the bindings. 

> I don't worry too much about an extra enum, but if anyone ever wanted this enum generated, either with LVGL or when generating bindings for some other library, you just blocked that option...

Yes, this is technically sadly a potential problem. But I don't know how to work around it with the current way pycparser is introducing this enum implicitly.

amirgon: Thanks for clarifying, but there is still something I don't understand.   
How comes including stdatomic.h in lv_mpy.c solved the issue, but including it indirectly (lv_mpy.c includes lvgl.h which includes stdatomic.h) doesn't work? Is there some include order issue here?

> This results into it appearing in the symbol file `gen_mpy.py` uses to generate the bindings.

I think that an enum appears in the generates bindings only if it's used, directly or indirectly, by some function that is processed by the script.
So even if the enum is defined (either in stdatomic.h or _fake_typedefs.h) it should be ignored by the script if LVGL doesn't use it (when atomics are disabled).

Sh3Rm4n: I've updated the PR, see [`bde6f21` (#255)](https://github.com/lvgl/lv_binding_micropython/pull/255/commits/bde6f21f37332e17647bf55142ed47eb9fd85ce6), which only contains the pycparser update itself. The LVGL is now set to a version, which does not use `stdatomic.h` directly.

I assume, because of the fact, that _fake_typedefs.h is included by **any** `std*` "fake"-header provided by pycparser, that it's very easy that the `memory_order` enum will also be included. 

AFAIK `gen_mpy.py` generates bindings for all symbols provided by lvgl header files, regardless of wether they are used by lvgl or not, but I can be wrong. Let's see what the CI of [`bde6f21` (#255)](https://github.com/lvgl/lv_binding_micropython/pull/255/commits/bde6f21f37332e17647bf55142ed47eb9fd85ce6) is saying.

amirgon: > AFAIK `gen_mpy.py` generates bindings for all symbols provided by lvgl header files, regardless of wether they are used by lvgl or not, but I can be wrong.

You are right about enums.  
Structs are generated only when used (therefore `atomic_flag` is not generated), but enums are always generated.

> So, a possible solution of mine was to exclude the `memory_order` enum: [`8572bb4` (#255)](https://github.com/lvgl/lv_binding_micropython/pull/255/commits/8572bb4671a61a7c74de1ec45616ec8a1de949c0).
> I know this might not be the best approach, but do you have something different in mind?

Ok got it. I don't see any other obvious solution.  
So let's proceed with your workaround on 8572bb4671a61a7c74de1ec45616ec8a1de949c0, I assume that under normal circumstances `memory_order` would never be needed on Micropython.

Issue #253, opened 2023-01-12
Arrays are converted to lists - cannot update existing arrays on structs


amirgon:
See: https://forum.lvgl.io/t/gradient-is-not-working-with-lv-draw-rect-dsc-t/10612

```python
rect_dsc.bg_grad.stops[0].color = lv.palette_main(lv.PALETTE.RED)
rect_dsc.bg_grad.stops[1].color = lv.palette_main(lv.PALETTE.BLUE)
```
does not work, because stops is converted to list and does not update the original array upon update.

A workaround can be to the the whole array:

```python
rect_dsc.bg_grad.stops = [
    lv.gradient_stop_t({'color': lv.palette_main(lv.PALETTE.RED)}),
    lv.gradient_stop_t({'color': lv.palette_main(lv.PALETTE.BLUE), 'frac':0xff})
]
```

PR #260, opened 2023-02-26
Hopefully the fix for #253

kdschlosser:
Making changes to a structure that is in an array does not populate to the array. This is hopefully the fix for #253 I have not tested this change yet and maybe someone that knows more about where the problem is exactly would be willing to test it.

### DIFF
--- a/gen/gen_mpy.py
+++ b/gen/gen_mpy.py
@@ -1969,9 +1969,12 @@ def try_generate_array_type(type_ast):
     arr_to_c_convertor_name = 'mp_arr_to_%s' % array_convertor_suffix
     arr_to_mp_convertor_name = 'mp_arr_from_%s' % array_convertor_suffix
 
-    if hasattr(type_ast.type, 'names'):
+    if isinstance(type_ast.type, c_ast.IdentifierType):
         for nme in type_ast.type.names:
-            if nme in lv_to_mp:
+            for strct in struct_typedefs:
+                if strct.declname == nme:
+                    break
+            else:
                 struct_type = False
                 break
         else:
--- a/gen/gen_mpy.py
+++ b/gen/gen_mpy.py
@@ -1968,6 +1968,17 @@ def try_generate_array_type(type_ast):
         replace('/','_div_')
     arr_to_c_convertor_name = 'mp_arr_to_%s' % array_convertor_suffix
     arr_to_mp_convertor_name = 'mp_arr_from_%s' % array_convertor_suffix
+
+    if hasattr(type_ast.type, 'names'):
+        for nme in type_ast.type.names:
+            if nme in lv_to_mp:
+                struct_type = False
+                break
+        else:
+            struct_type = True
+    else:
+        struct_type = False
+
     print((('''
 /*
  * Array convertors for {arr_name}
@@ -2013,7 +2024,7 @@ GENMPY_UNUSED STATIC mp_obj_t {arr_to_mp_convertor_name}({qualified_type} *arr)
         qualified_ptr_type = qualified_element_ptr_type,
         check_dim = '//TODO check dim!' if dim else '',
         mp_to_lv_convertor = mp_to_lv[element_type],
-        lv_to_mp_convertor = lv_to_mp[element_type],
+        lv_to_mp_convertor = lv_to_mp[element_type_ptr] if struct_type and dim else lv_to_mp[element_type],
         mp_to_lv_ptr_convertor = mp_to_lv[element_type_ptr],
         lv_to_mp_ptr_convertor = lv_to_mp[element_type_ptr],
         dim = dim if dim else 1,


gen/gen_mpy.py
```diff
@@ -1968,6 +1968,17 @@ def try_generate_array_type(type_ast):
         replace('/','_div_')
     arr_to_c_convertor_name = 'mp_arr_to_%s' % array_convertor_suffix
     arr_to_mp_convertor_name = 'mp_arr_from_%s' % array_convertor_suffix
+
+    if hasattr(type_ast.type, 'names'):
```
amirgon, line 1972: Is this really the best way to test for a struct type?  
There's already a function for it actually:

https://github.com/lvgl/lv_binding_micropython/blob/f960502a3260f795f21bfa90b2e7d7199430cf88/gen/gen_mpy.py#L287-L288

gen/gen_mpy.py
```diff
@@ -2013,7 +2024,7 @@ def try_generate_array_type(type_ast):
         qualified_ptr_type = qualified_element_ptr_type,
         check_dim = '//TODO check dim!' if dim else '',
         mp_to_lv_convertor = mp_to_lv[element_type],
-        lv_to_mp_convertor = lv_to_mp[element_type],
+        lv_to_mp_convertor = lv_to_mp[element_type_ptr] if struct_type and dim else lv_to_mp[element_type],
```
amirgon, line 2027: `try_generate_array_type` is used in a wider context. For example, when returning a struct as a return value from a function. In such case, it's wrong to pass the struct by reference.  
So I suspect that just updating `lv_to_mp_convertor` is too general and could be wrong in certain cases.

amirgon (changes requested):
Hi @kdschlosser , thank you for trying to help with this fix.
See my comments below.

gen/gen_mpy.py
kdschlosser: the reason I am checking for the names attribute is because there are 2 different class instances that come through that function. One is IdentifierType which has the names attribute. the Other I cannot remember off the top of my head but it does not have the names attribute.  The type that comes through is not a c_ast.Struct or c_ast.Union instance It is an array and that array is filled with something. I am doing a check to make sure it is not filled with integers or bools as these cause problems when you set a point for each individual item in the array. We only want to use pointers for Structures and Unions that are in the array.

This is what we are looking for in terms of what `type_ast` should be
```
TypeDecl(declname='stops', quals=[], type=IdentifierType(names=['lv_gradient_stop_t' ))
```

and not this

```
TypeDecl(declname='inv_area_joined', quals=[], type=IdentifierType(names=['uint8_t'])
```

The only things that get processed by that function are
```
DIM: 2
TYPE: TypeDecl(declname='stops', quals=[], type=IdentifierType(names=['lv_gradient_stop_t']))

DIM: None
TYPE: TypeDecl(declname='points', quals=['const'], type=IdentifierType(names=['lv_point_t']))

DIM: 32
TYPE: TypeDecl(declname='inv_areas', quals=[], type=IdentifierType(names=['lv_area_t']))

DIM: 32
TYPE: TypeDecl(declname='inv_area_joined', quals=[], type=IdentifierType(names=['uint8_t']))

DIM: None
TYPE: TypeDecl(declname='col_dsc', quals=['const'], type=IdentifierType(names=['lv_coord_t']))

DIM: None
TYPE: TypeDecl(declname='props', quals=['const'], type=IdentifierType(names=['lv_style_prop_t']))
```

and of those only these ones actually end up having any changed made to them
```
DIM: 2
TYPE: TypeDecl(declname='stops', quals=[], type=IdentifierType(names=['lv_gradient_stop_t']))

DIM: 32
TYPE: TypeDecl(declname='inv_areas', quals=[], type=IdentifierType(names=['lv_area_t']))
```

that is because the others have a dim of None or the contents of the array is not a structure/union


and on the esp32 side of things
```
DIM: 6
TYPE: TypeDecl(declname='mac', quals=[], type=IdentifierType(names=['uint8_t']))

DIM: 4
TYPE: TypeDecl(declname='addr', quals=[], type=IdentifierType(names=['u32_t']))

DIM: 4
TYPE: TypeDecl(declname='tx_data', quals=[], type=IdentifierType(names=['uint8_t']))

DIM: 4
TYPE: TypeDecl(declname='client_ip', quals=[], type=IdentifierType(names=['u8_t']))

DIM: None
TYPE: TypeDecl(declname='nva', quals=['const'], type=Struct(name='sh2lib_nv', decls=None))

DIM: 10
TYPE: TypeDecl(declname='sta', quals=[], type=IdentifierType(names=['esp_netif_sta_info_t']))

DIM: None
TYPE: TypeDecl(declname='if_ip6', quals=[], type=IdentifierType(names=['esp_ip6_addr_t']))

DIM: None
TYPE: TypeDecl(declname='txt', quals=[], type=IdentifierType(names=['mdns_txt_item_t']))
```

this is what ends up getting changed
```
DIM: 10
TYPE: TypeDecl(declname='sta', quals=[], type=IdentifierType(names=['esp_netif_sta_info_t']))
```

gen/gen_mpy.py
kdschlosser: see the reply in the previous comment and you will see that the only thing that are effected is a really small list and what is being changed are the things that do need to be changed.

gen/gen_mpy.py
kdschlosser: I am not sure wh `TypeDecl(declname='nva', quals=['const'], type=Struct(name='sh2lib_nv', decls=None))` is even making it to that function as it is a Structure and not an array. That is the reason why I am checking for the `names` attribute.

kdschlosser: There also seems to be an apparent problem somewhere because `inv_areas` does not have 32 dimensions and `stops` does not have 2 dimensions. They are both single dimension arrays and one has 32 elements in that array and the other has 2 elements in that array. a multidimensional array is a nested array as you know. in Python it would be 
```
some_list = [[0]] 
```

That is a 2 dimension array and there is a single element in the second dimensions. I do not believe that you can mix types and or dimensions in C code. meaning you cannot do this

```
some_list = ['hellow world', [0, 0, 0]]
```

so I am not really sure as to why pycparser has list of type names for the array

kdschlosser: as another note and it might be worth opening an issue for with pycparser. I think there should be a variable in `IdentifierType` that holds the "type" instance for what is actually being stored in the array instead of storing the name of the type. this way there would end up being a really solid way of being able to tell what the type is that is being stored in the array.Right now this kind of C code would much the whole thing up.

```
typedef uint8_t my_uint8_t;

my_uint8_t *my_array[15];
```

that is pseudo code and the syntax is probably wrong but you get the general idea.

What would end up happening is you would end up with pycparser doing this.

```
TypeDecl(declname='my_array', quals=[], type=IdentifierType(names=['my_uint8_t']))
```

and there is no concrete way of knowing if my_uint8_t is a struct, union or anything else. But if what was stored was the original declaration instead of just a name it would work out better.

That being said. in the gen_mpy.py script is there a name to pycparser type mapping?? if there is then maybe that can be used as the best way to nail down if it is a structure or a union.

PR #278, opened 2023-06-09
open librt from librt.so.1

starstar222:
Adapt to embedded board

### DIFF
--- a/driver/linux/lv_timer.py
+++ b/driver/linux/lv_timer.py
@@ -17,7 +17,7 @@ import sys
 
 libc = ffi.open("libc.so.6")
 try:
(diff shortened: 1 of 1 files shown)


amirgon:
Hi @liuyunrui123 
Could you clarify this a little?

In general, `librt.so` should be a symbolic link to the specific version of the library (for example, `librt.so.1`). Why is it missing on your platform? Perhaps something is not installed correctly on your side?  

Your change would break on platforms where a higher versions of `librt.so` is installed (for example, if `librt.so.2` is installed in the system instead of `librt.so.1`), that's why it might make more sense to not use the specific version of the library when loading it.

PR #335, opened 2024-03-04
fixes empty structures being created

kdschlosser:
The forward declarations has broken things like event callbacks due to structures not being populated with their fields properly This fixes that issue y searching for the forward declarations and their typedefs and making a single object to represent the structure with all the proper fields.

### DIFF
--- a/gen/gen_mpy.py
+++ b/gen/gen_mpy.py
@@ -293,6 +293,58 @@ parser = c_parser.CParser()
 gen = c_generator.CGenerator()
 ast = parser.parse(s, filename='<none>')
 
+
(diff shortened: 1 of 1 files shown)


kisvegabor:
The binding is also updated in `lv_micropython`.

kdschlosser:
OK make sure the fix allows the test to pass on the PR I made to LVGL for the test update,

kisvegabor:
But the CI in lvgl is still failing. 🙁

kdschlosser:
I know, I am working on it. I can't sit that long at the computer.

PR #349, opened 2024-07-19
feat(gen): accept a JSON from the LVGL JSON generator

liamHowatt:
For lvgl/lvgl#6068

This is a backwards-compatible change that allows gen_mpy.py to accept a JSON from the LVGL JSON generator.

Currently it only uses the JSON to find the `user_data` member of private structs.


liamHowatt:
Blocking lvgl/lv_micropython#84

kdschlosser:
for a minute there you had me thinking that the gen_mpy script had been rewritten to use the JSON output instead of Pycparser directly.

kisvegabor:
I would be ok with making the MP binding fully work based on the JSON. But let's proceed step by step. :)

PR #350, opened 2024-07-19
fix(gen): Use LVGL JSON in CMake based builds too - RP2040 and ESP32

liamHowatt:
For lvgl/lvgl#6068

- Use LVGL JSON in CMake based builds, i.e. RP2040 and ESP32.
- Use updated name for `lv_indev_t` in STM32 source file.


kisvegabor:
Awesome!

PR #383, opened 2025-05-18
(fix) lvgl deinit on soft-reset

### LINKED ISSUE #343: Soft reboot failure
I wasn't sure whether to report this as an issue or not, so feel free to immediately close it, I just wanted to document this for others to find if they were experiencing the same issues as me.

TLDR: using the default `lv_conf.h` also present in this repository, I failed to have soft-reboots working properly, but I found a solution.

The first hurdle, getting to a clean state, was fairly easily taken by implementing finalizers and using the `deinit()` function. After that, I could completely tear down and rebuild my setup by manually calling the functions in REPL multiple times in a row and it would keep working.

The second problem wasn't so easy to debug: once I soft rebooted, the finalizers were called but it would always crash with a MemoryError. After a lot of scratching my head and a deep-dive into the memory allocation code of both LVGL and MicroPython I found finally that during `deinit()` the LVGL code is doing calls to `realloc` which are failing because the garbage collector doesn't want to hand out memory anymore. Whether this is a problem to be solved in LVGL or in MicroPython I'm not sure at this point.

The solution finally was pretty simple: stop using the `LV_STDLIB_MICROPYTHON` malloc implementation :shrug: 

Here you can see how I solved it in my implementation:
https://github.com/MathyV/lvgl_esp32_mpy/commit/e84196d20415162422b7ead386934419384a66d6

I don't know if this is the best (temporary?) solution but so far I haven't experienced any issues with it yet.

PS: after fixing this I also had issues with `deinit()` of `machine.SPI` in the ESP32 MicroPython implementation but that is definitely out of scope here, still, here's how I fixed it as an FYI:
https://github.com/MathyV/lvgl_esp32_mpy/commit/7330fcc6bdc23e84a8e1bb8db0970650338f5940

Carglglz:
Include `lv_conf.h` in `mpconfigboard.h` so lvgl deinit is called on soft-reset, so calling `lvgl.deinit()` is no longer necessary.

This closes #343

### DIFF
--- a/lv_conf.h
+++ b/lv_conf.h
@@ -306,6 +306,10 @@ extern void mp_lv_deinit_gc();
 #define LV_GC_INIT() mp_lv_init_gc()
 #define LV_GC_DEINIT() mp_lv_deinit_gc()
 
+// include lv_conf.h in "mpconfigboard.h" for this to take effect.
+extern void lv_deinit();
+#define MICROPY_PORT_DEINIT_FUNC lv_deinit()
+
 #define LV_ENABLE_GLOBAL_CUSTOM 1
 #if LV_ENABLE_GLOBAL_CUSTOM
     extern void *mp_lv_roots;


PGNetHun (approved):

Carglglz:
This seems to be a better solution, I've tested stm32 port and it seems to work, esp32 will be next, so do not merge this yet...

Also I had to update CI make command for esp32, I guess you added `USER_C_MODULES` in `mpconfigboard.cmake` for `ESP32_GENERIC_S3` board in `lv_micropython`?

Carglglz:
@PGNetHun I've just tested the esp32 port and it seems to be working as expected, so it is ready for review/merge now 👍🏼

.github/workflows/micropython_lvgl_ci.yml
```diff
@@ -83,12 +83,14 @@ jobs:
         make -j $(nproc) -C ports/rp2 BOARD=RPI_PICO USER_C_MODULES=../../user_modules/lv_binding_micropython/micropython.cmake
 
     # ESP32 port
+        # USER_C_MODULES defined in mpconfigboard.cmake
```
PGNetHun, line 86: Please remove this line, it is incorrect.
In ESP32 port `USER_C_MODULES` is defined in `esp32_common.cmake` file:
https://github.com/lvgl/lv_micropython/blob/master/ports/esp32/esp32_common.cmake#L14

lv_conf.h
```diff
@@ -434,6 +434,9 @@ extern void mp_lv_deinit_gc();
 #define LV_GC_INIT() mp_lv_init_gc()
 #define LV_GC_DEINIT() mp_lv_deinit_gc()
 
+// To enable deinit on soft-reset add in mpconfigboard.h
```
PGNetHun, line 437: What if you define deinit here with a new `LV_` config?
Like:

```
#define LV_ENABLE_DEINIT_ON_SOFT_RESET 1
#ifdef LV_ENABLE_DEINIT_ON_SOFT_RESET
   extern void mp_deinit_lvgl_mod();
   #define MICROPY_PORT_DEINIT_FUNC mp_deinit_lvgl_mod()
#endif
```

include/lv_mp_port_soft_reset.h
PGNetHun, line 1: No need for this file if deinit is defined in `lv_conf.h` like mentioned in other comment

PGNetHun (changes requested):
Added some comments, suggestion where to define to call deinit function.
What do you think?

include/lv_mp_port_soft_reset.h
Carglglz: > Well, I think we should define that MICROPY_PORT_DEINIT_FUNC macro in other place, not in lv_conf.h and then explicitly include that config file in mpconfigboard.h.

One way or the other, so which one ?

PR #411, opened 2026-07-20
fix(gen): support MicroPython's renamed mp_obj_int_to_bytes() API

andrewleech:
### Summary

MicroPython commit [e00daa3a](https://github.com/micropython/micropython/commit/e00daa3a) ("py/binary,objint: Add overflow checks and `int.to_bytes(signed=True)`") renamed `mp_obj_int_to_bytes_impl()` to `mp_obj_int_to_bytes()` and changed its signature, adding `is_signed` and `overflow_check` parameters. The generated binding calls this helper in `mp_obj_get_ull()`, so building against MicroPython 1.29 or later fails to link with `undefined reference to 'mp_obj_int_to_bytes_impl'`.

This guards the call on `MICROPY_VERSION >= MICROPY_MAKE_VERSION(1, 29, 0)`, calling the new API on 1.29+ and the old one on earlier versions. The reverse helper `mp_obj_int_from_bytes_impl()` was not renamed and is unchanged.

lv_micropython merges the upstream MicroPython master that carries the rename, so it needs the same fix: lvgl/lv_micropython#106.

### Testing

Built the binding into a MicroPython firmware for an i.MX RT1176 board (mimxrt port, arm-none-eabi) against a 1.29 MicroPython: it links, and `mp_obj_get_ull()` compiles the `mp_obj_int_to_bytes()` call. The pre-1.29 branch of the guard was checked by preprocessing against a 1.24 header, which selects the old call.

I did not test the unix port.

### Generative AI

I used generative AI tools when creating this PR, but a human has checked the code and is responsible for the code and the description above.

### DIFF
--- a/gen/gen_mpy.py
+++ b/gen/gen_mpy.py
@@ -1832,7 +1832,13 @@ static unsigned long long mp_obj_get_ull(mp_obj_t obj)
 
     unsigned long long val = 0;
     bool big_endian = !(__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__);
+    // mp_obj_int_to_bytes() replaced mp_obj_int_to_bytes_impl() in MicroPython
+    // 1.29 (commit e00daa3a), adding is_signed and overflow_check parameters.
+#if MICROPY_VERSION >= MICROPY_MAKE_VERSION(1, 29, 0)
+    mp_obj_int_to_bytes(obj, sizeof(val), (byte*)&val, big_endian, false, false);
+#else
     mp_obj_int_to_bytes_impl(obj, big_endian, sizeof(val), (byte*)&val);
+#endif
     return val;
 }
 
--- a/gen/lv_mpy_example.c
+++ b/gen/lv_mpy_example.c
@@ -778,7 +778,13 @@ STATIC unsigned long long mp_obj_get_ull(mp_obj_t obj)
 
     unsigned long long val = 0;
     bool big_endian = !(__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__);
+    // mp_obj_int_to_bytes() replaced mp_obj_int_to_bytes_impl() in MicroPython
+    // 1.29 (commit e00daa3a), adding is_signed and overflow_check parameters.
+#if MICROPY_VERSION >= MICROPY_MAKE_VERSION(1, 29, 0)
+    mp_obj_int_to_bytes(obj, sizeof(val), (byte*)&val, big_endian, false, false);
+#else
     mp_obj_int_to_bytes_impl(obj, big_endian, sizeof(val), (byte*)&val);
+#endif
     return val;
 }
 


o-murphy:
Confirming independent testing on the unix port (you noted you hadn't tried it): rebuilt `ports/unix` against current MicroPython master (`5aaa6d17`) via `USER_C_MODULES` with this exact fix applied - builds clean, `import lvgl` works, and a smoke test creating `obj`/`button`/`label` with `add_event_cb`/`send_event` runs correctly end to end. The `#if MICROPY_VERSION >= MICROPY_MAKE_VERSION(1, 29, 0)` guard is the right call - I'd independently landed on the same `mp_obj_int_to_bytes(obj, sizeof(val), (byte*)&val, big_endian, false, false)` replacement call but without the backward-compat guard, so this version is the better fix.

Side note: this is very likely also why the `tests/api` CI jobs (`basic.py`, `basic_indev.py`, `basic_slider.py`) have been failing on `master` since [run 28447459670](https://github.com/lvgl/lv_binding_micropython/actions/runs/28447459670/job/84300491736) (2026-06-30, six days after the breaking MicroPython commit) - same root cause, worth a rebase/rerun once this merges to confirm.

o-murphy:
@andrewleech @pi-anl 

Confirming this also fixes the unix port, since the description notes it wasn't tested there.

Applied the same guard to `gen/gen_mpy.py` and `gen/lv_mpy_example.c` and built `ports/unix` as a `USER_C_MODULES` usermod against a current MicroPython checkout (`sys.implementation.version == (1, 29, 0, 'preview')`), which is right at the `MICROPY_VERSION >= MICROPY_MAKE_VERSION(1, 29, 0)` boundary this PR adds, so the new `mp_obj_int_to_bytes()` branch is the one actually compiled and exercised, not the old one.

Built and ran it repeatedly (dozens of rebuilds over the last day), including live LVGL UI code -- buttons, labels, style colors, indev state -- that routes through `mp_obj_get_ull()` in the generated binding. No link errors, no runtime issues.

So: confirmed working on unix in addition to the mimxrt testing already in the description.

o-murphy:
Confirming this also fixes the unix port (the description notes it wasn't tested there): built `ports/unix` as a `USER_C_MODULES` usermod against a current MicroPython checkout (`sys.implementation.version == (1, 29, 0, 'preview')`), right at the `MICROPY_VERSION >= MICROPY_MAKE_VERSION(1, 29, 0)` boundary this PR adds, so the new `mp_obj_int_to_bytes()` branch is the one actually compiled and exercised. Built and ran it repeatedly over the last few days, including real LVGL UI code (buttons, labels, style colors, indev state) that routes through `mp_obj_get_ull()` in the generated binding - no link errors, no runtime issues.

Worth calling out how live this bug currently is: CI on `master` is failing this exact way on **all four ports** right now - [run 30154151496](https://github.com/lvgl/lv_binding_micropython/actions/runs/30154151496), 4 days ago, unix/stm32/rp2/esp32 all hit `implicit declaration of function 'mp_obj_int_to_bytes'; did you mean 'mp_obj_int_to_bytes_impl'?`. That run was even built against [7ee268e](https://github.com/lvgl/lv_binding_micropython/commit/7ee268e176e6a99d2c56fbf8be30c416f6159416) ("fix: update mp_obj_int_to_bytes implementation for correct byte conversion"), a different attempt at this same problem - it doesn't actually resolve it, same error either way. None of this shows up as red because the workflow sets `continue-on-error: true` per job.

We've got a follow-up PR (#413) building on top of this one - adds LVGL's native Wayland driver, and separately fixes a regression that's kept every desktop window driver (SDL included) out of the Python bindings since May. Rebasing that once this lands.

